PR #122, opened 2022-03-17
Idea cards small

LetsPoke:
- created small idea card component
- added some content to the pages
- changed browser titel and image


hms-app/src/TestIdeaData.js
```diff
@@ -0,0 +1,18 @@
...
+        description:"just a better idea ",
+        favNumber: 8
+    },
+    {
+        titel:"idea 3",
+        description:"absolute the best idea in the world ",
+        favNumber: 5
+    },
+
+]
```
verzano, line 18: it's a weird old fashioned thing but when possible files should always end in a newline

LetsPoke: thanks for the hint

jakepanter (approved):

PR #161, opened 2022-03-25
Adding test setup to the HMS API

verzano:
Added the setup for Jest and added a fake test as an example


LetsPoke (commented):
looks good to me but Jakob should merge instead of me

jakepanter (approved):

PR #254, opened 2022-04-19
185 connect frontend with backend

### LINKED ISSUE #190: add functionality to the archive
- create a dropdown with the hackathon names and years
- display some details for the selected hackathon (title, date,...)
- list all ideas that belongs to the hackathon

LetsPoke:
- added dropdown for hackweek select in archvie
- display some details for selected hackweek
- loads ideas that are connected to the selected hackweek

closes #190


hms-app/src/actions/GetBackendData.tsx
```diff
@@ -0,0 +1,52 @@
+const api_id = '6cyvp5ukhc'
+
+export const getListOfHackathons = (endpoint: any) => {
+  return fetch(
+    `${'http://localhost:4566/restapis/'}${api_id}${'/local/_user_request_'}/${endpoint}`,
```
verzano, line 5: so since you aren't executing any JS code or referencing variables you don't need the `${...}` around things like `http://local...` or `/loca/_user...`.

Same thing goes for the other places like that

hms-app/src/actions/GetBackendData.tsx
```diff
@@ -0,0 +1,52 @@
+const api_id = '6cyvp5ukhc'
+
+export const getListOfHackathons = (endpoint: any) => {
```
verzano, line 3: the name here implies that that `endpoint` will always be `hackathons` so you likely don't need to pass in a value.  You should also specify a return type here :thinking:

hms-app/src/actions/GetBackendData.tsx
```diff
@@ -0,0 +1,52 @@
...
+      },
+    }
+  )
+    .then((response) => {
+      return response.json()
+    })
+    .catch((err) => console.log(err))
+}
+
+export const getIdeaDetails = (endpoint: any) => {
```
verzano, line 37: the argument could have a better name, something like `id` or `ideaId` so it is more obvious what should be passed in

hms-app/src/actions/GetBackendData.tsx
```diff
@@ -0,0 +1,52 @@
...
+      },
+    }
+  )
+    .then((response) => {
+      return response.json()
+    })
+    .catch((err) => console.log(err))
+}
+
+export const getHackathonDetails = (endpoint: any) => {
```
verzano, line 20: same thing here about the argument name, `id` or `hackathonId`

hms-app/src/common/types.tsx
```diff
@@ -1,20 +1,50 @@
...
+
+export type IdeaPreviews = {
+  id: string
+  title: string
+}
+
 export type Idea = {
   id: string
+  owner: Owner[]
+  hackathon: HackathonPreview[]
```
verzano, line 14: this will never be arrays

hms-app/src/common/types.tsx
```diff
@@ -1,20 +1,50 @@
...
-  author?: User
-  reason: string
   problem: string
   goal: string
-  skills: string[]
-  category: string
-  participants: User[]
-  aws: boolean
+  requiredSkills: Skill[]
+  category: Category[]
```
verzano, line 21: same here, will always just be a single value not an array

hms-app/src/common/types.tsx
```diff
@@ -1,20 +1,50 @@
...
+  id: string
   name: string
-  email?: string
-  role?: string
-  avatar?: string
+}
+
+export type Owner = {
+  id: string
+  user: User[]
```
verzano, line 37: should only be a single value not an array

hms-app/src/common/types.tsx
```diff
@@ -1,20 +1,50 @@
...
+}
+
+export type Participant = {
+  id: string
+  user: User[]
```
verzano, line 42: same about the single value

hms-app/src/components/HackathonDetails.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+type IProps = {
+  hackathonID: string
+}
+
+export default function HackathonDetails(props: IProps) {
+  const { hackathonID } = props
+  const [hackathonData, setHackathonData] = useState({
+    errorhackathonData: false,
+    isLoadinghackathonData: true,
```
verzano, line 13: You would need to fix this everywhere they are used, but the names should be proper camel-case:

```suggestion
    errorHackathonData: false,
    isLoadingHackathonData: true,
```

hms-app/src/components/HackathonDetails.tsx
```diff
@@ -0,0 +1,89 @@
...
+    title,
+    startDate,
+    endDate,
+    participantIds,
+    categoryIds,
+    ideas,
+  } = hackathonData
+
+  const loadSelectedHackathon = () => {
+    getHackathonDetails(hackathonID.toString()).then(
```
verzano, line 34: since you define it's type as `string` already` don't need to call `toString()`
```suggestion
    getHackathonDetails(hackathonID).then(
```

hms-app/src/components/HackathonDetails.tsx
```diff
@@ -0,0 +1,89 @@
...
+export default function HackathonDetails(props: IProps) {
+  const { hackathonID } = props
+  const [hackathonData, setHackathonData] = useState({
+    errorhackathonData: false,
+    isLoadinghackathonData: true,
+    title: 'string',
+    startDate: '',
+    endDate: '',
+    participantIds: [],
+    categoryIds: [],
```
verzano, line 18: Since we now return the `XxxPreviewResponse` for everything, these names are probably a bit more accurate:
```suggestion
    participants: [],
    categories: [],
```

hms-app/src/components/HackathonDetails.tsx
```diff
@@ -0,0 +1,89 @@
...
+        </div>
+      )}
+
+      {startDate && (
+        <div>
+          <p>selected ID: {hackathonID}</p>
+          <h2>Title: {hackathonData.title}</h2>
+          <h2>
+            Date from: {hackathonData.startDate.slice(0, 10)} to:{' '}
+            {hackathonData.endDate.slice(0, 10)}
```
verzano, line 81: startDate and endDate should always be nicely formatted ISO date strings, you could easily parse them be calling `new Date(startDate)` and then print them formatted as you choose.

hms-app/src/components/IdeaCardFoldable.tsx
```diff
@@ -12,14 +12,38 @@ import {
...
+import { getIdeaDetails } from '../actions/GetBackendData'
 
 type IProps = {
-  idea: Idea
+  ideaPreview: IdeaPreviews
   index: number
   type: string
 }
 
+type Idea = {
```
verzano, line 31: you defined an idea type in your `types.tsx` file, why not use that here?

verzano: I guess you also have the `error` and `loading` state flags, but you could do something like:
```
type IdeaCardState = {
  errorIdeaData: boolean
  isLoadingIdeaData: boolean
  idea: Idea
}
```

hms-app/src/components/IdeaCardFoldable.tsx
```diff
@@ -50,136 +74,191 @@ export default function IdeaCardFoldable(props: IProps) {
...
-      {skill}
-    </Badge>
-  ))
-
-  const participantAvatars = idea.participants.map((participant) => (
-    <Avatar key={participant.name} src={participant.avatar} />
-  ))
+  useEffect(() => {
+    loadIdeaDetails()
+  }, [])
```
verzano, line 145: technically if the `ideaPreview` variable changed, this should reload.  But I don't know if react will reuse components like that or if it just makes new ones :thinking:.  That is to say this is most likely fine with just `[]`, more I'm just thinking out loud... or in text on the internet

hms-app/src/pages/Archive.tsx
```diff
@@ -1,18 +1,74 @@
...
+interface HackathonPreview {
+  id: string
+  title: string
+}
 
 export default function Archive() {
+  const [selectedHackweek, setSelectedHackweek] = useState('')
+  const [hackathonList, setHackathonList] = useState({
+    errorhackathonList: false,
+    isLoadinghackathonList: true,
```
verzano, line 15: Would also need to fix this everywhere
```suggestion
    errorHackathonList: false,
    isLoadingHackathonList: true,
```

hms-app/src/pages/Archive.tsx
```diff
@@ -1,18 +1,74 @@
...
-      <IdeaCardList
-        ideas={ideaData as Idea[]}
-        columnSize={6}
-        type={'archive'}
-      />
+      {isLoadinghackathonList && <div>hackathon select is loading...</div>}
+      {!isLoadinghackathonList && (
+        <div style={{ border: '1px solid red', width: 250 }}>
+          <Select
+            placeholder={'select a Hackweek'}
```
verzano, line 60: Probably says `Hackweek` in other places as well

```suggestion
            placeholder={'select a Hackathon'}
```

hms-app/src/pages/Archive.tsx
```diff
@@ -1,18 +1,74 @@
...
-import { Idea } from '../common/types'
+import React, { useEffect, useState } from 'react'
+import { getListOfHackathons } from '../actions/GetBackendData'
+import HackathonDetails from '../components/HackathonDetails'
+import { Select } from '@mantine/core'
+
+interface HackathonPreview {
+  id: string
+  title: string
+}
```
verzano, line 9: Could probably define this in the `types.tsx` file as well :thinking:

hms-app/src/components/HackathonDetails.tsx
verzano: all of this data could be defined in a single type in your `types.tsx` file too

hms-app/src/common/types.tsx
```diff
@@ -1,20 +1,50 @@
...
+  creationDate: string
 }
 
-export type User = {
+export type Category = {
+  id: string
+  title: string
+}
+
+export type Skill = {
```
verzano, line 30: you might want to specify this (and many of the other things in here) as `XxxPreview` so that it is obvious that they don't contain all of the possible data

hms-app/src/common/types.tsx
LetsPoke: sure about that? dont we get an array with skill objects?

hms-app/src/actions/GetBackendData.tsx
```diff
@@ -0,0 +1,52 @@
...
+      },
+    }
+  )
+    .then((response) => {
+      return response.json()
+    })
+    .catch((err) => console.log(err))
+}
+
+export const getIdeaDetails = (ideaID: string) => {
```
verzano, line 37:
```suggestion
export const getIdeaDetails = (ideaID: string): Idea => {
```

hms-app/src/actions/GetBackendData.tsx
```diff
@@ -0,0 +1,52 @@
...
+    {
+      method: 'GET',
+      headers: {
+        Accept: 'application/json',
+        'Content-Type': 'application/json',
+      },
+    }
+  )
+    .then((response) => {
+      return response.json()
```
verzano, line 49:
```suggestion
      return  new Idea(....)
```

hms-app/src/components/IdeaCardFoldable.tsx
```diff
@@ -50,136 +51,175 @@ export default function IdeaCardFoldable(props: IProps) {
...
+    title: 'string',
+    description: 'string',
+    problem: 'string',
+    goal: 'string',
+    requiredSkills: [],
+    category: undefined,
+    creationDate: 'string',
+  } as Idea)
+
+  const idea: Idea = ideaData
```
verzano, line 70: you actually don't need this line.  You could now access ideaData directly (or rename it to idea)

verzano (approved):

PR #262, opened 2022-04-21
256 ideaportal functionality

### LINKED ISSUE #170: text search differs between upper and lowercase
text search should not differ between upper and lowercase

### LINKED ISSUE #256: IdeaPortal functionality
- display all ideas of the current hackathon
- add the search function to this new list

LetsPoke:
- added base funtions to the idea portal 
- fixed text search
- updated some problems in the archieve

closes #256 
closes #170


hms-app/src/components/HackathonDetails.tsx
```diff
@@ -46,8 +62,60 @@ export default function HackathonDetails(props: IProps) {
 
   useEffect(() => {
     loadSelectedHackathon()
+    setRelevantIdeaList([])
```
verzano, line 65: I'm a little worried you built a bit of a race condition in here :thinking: I'm not sure if all these useEffect hooks execute at that same time or one after another... if it is the same time you call setRelevantIdeaList here and in the useEffect that is triggered by changes to ideaData...

Actually looking at this more you never seem to use ideaData by itself so you could remove it

hms-app/src/components/IdeaCardFoldable.tsx
```diff
@@ -16,7 +16,7 @@ import { Idea, IdeaPreview } from '../common/types'
 import { getIdeaDetails } from '../actions/GetBackendData'
 
 type IProps = {
-  ideaPreview: IdeaPreview
+  ideas: Idea
```
verzano, line 19: Since it is just singular
```suggestion
  idea: Idea
```

PR #301, opened 2022-04-28
258 prepare test hackathon

LetsPoke:
added a lot of admin tools


hms-app/src/actions/ParticipantActions.tsx
```diff
@@ -0,0 +1,42 @@
...
+    .then((response) => {
+      return response.json()
+    })
+    .catch((err) => console.log(err))
+}
+
+export const createHackathonParticipant = (props: {
+  userId: string
+  hackathonId: string
+}) => {
```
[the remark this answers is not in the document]
verzano: the name of this function could be shortened to just `createParticipant` you don't prefix anything with `Hackathon` anywhere else that I can see. It will shorten overall line length this way

hms-app/src/components/AllHackathonList.tsx
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  const allHackathons = hackathonList.hackathons.map((hackathon, index) => [
+    <SimpleGrid
+      cols={2}
+      breakpoints={[{ maxWidth: 'sm', cols: 1 }]}
+      className={index % 2 ? classes.list : classes.list2}
+    >
+      <li>title: {hackathon.title}</li>
+      ID: {hackathon.id}
```
verzano, line 69: I would think this would be a bit more correct like:
```suggestion
      <div>title: {hackathon.title}</div>
      <div>ID: {hackathon.id}</div>
```

hms-app/src/components/AllHackathonList.tsx
```diff
@@ -0,0 +1,102 @@
...
+    })
+    loadHackathons()
+  }
+
+  return (
+    <>
+      <Card withBorder radius="md" p="md" className={classes.card}>
+        <Card.Section className={classes.section}>
+          <Group position="left" mt="xl">
+            <Button onClick={refreshList}>Refresh list</Button>
```
verzano, line 90: to prevent people clicking multiple times and triggering a bunch of requests you could do something like:
```suggestion
            {!hackathonlist.isLoadingHackathonList && <Button onClick={refreshList}>Refresh list</Button>}
```

hms-app/src/components/AllHackathonList.tsx
```diff
@@ -0,0 +1,102 @@
...
+  },
+}))
+
+function AllHackathonList() {
+  const { classes } = useStyles()
+  const [hackathonList, setHackathonList] = useState({
+    errorHackathonList: false,
+    isLoadingHackathonList: true,
+    hackathons: [] as HackathonPreview[],
+  })
```
verzano, line 40: so the anal-retentive part of me would say that this is more than just the `hackathonList`.  I can't remember completely but I think in my React class they said things like this would generally get split apart.  So you would have something more like;
```
const [isError, setIsError] = useState(false);
const [isLoading, setIsLoading] = useState(false);
const [hackathonsList, setHackathonsList] = useState([] as HackathonPreview[]);
```

then it becomes a bit easier if you really only need to change one of these things when something happens and you need not do so much `...hackathonList` stuff.  But it isn't anything to change right now.  Maybe play with it a bit and see which you like more or is easier to understand when reading the code.

hms-app/src/components/AllHackathonList.tsx
```diff
@@ -0,0 +1,102 @@
...
+
+  const allHackathons = hackathonList.hackathons.map((hackathon, index) => [
+    <SimpleGrid
+      cols={2}
+      breakpoints={[{ maxWidth: 'sm', cols: 1 }]}
+      className={index % 2 ? classes.list : classes.list2}
+    >
+      <li>title: {hackathon.title}</li>
+      ID: {hackathon.id}
+    </SimpleGrid>,
```
verzano, line 70: I think for these to fit inside a `<ol>` tag properly the whole thing should be wrapped in `<li>` tags.  The outer tag should also specify a key to make React happy like so:
```
<li key={hackathon.id}>
...
</li>
```

they talk about keys in lists in this [article](https://reactjs.org/docs/lists-and-keys.html)

hms-app/src/components/HackathonDetails.tsx
```diff
@@ -116,6 +217,45 @@ export default function HackathonDetails(props: IProps) {
     }
   }, [ideaData])
 
+  const participantsPreviewData = hackathonData.participants?.map(
+    (participant, index) => (
+      <SimpleGrid
```
verzano, line 222: same thing here (and in the other lists) about them being wrapped in `<li>` and having a `key` set

hms-app/src/components/HackathonDetails.tsx
```diff
@@ -116,6 +217,45 @@ export default function HackathonDetails(props: IProps) {
...
+        cols={2}
+        breakpoints={[{ maxWidth: 'sm', cols: 1 }]}
+        className={index % 2 ? classes.list : classes.list2}
+      >
+        <li>
+          Name: {participant.user.firstName} {participant.user.lastName}
+        </li>
+        ID: {participant.user.id}
```
verzano, line 230: and here about them being `<div>` elements

hms-app/src/components/AllHackathonList.tsx
```diff
@@ -0,0 +1,102 @@
...
+    <>
+      <Card withBorder radius="md" p="md" className={classes.card}>
+        <Card.Section className={classes.section}>
+          <Group position="left" mt="xl">
+            <Button onClick={refreshList}>Refresh list</Button>
+            {hackathonList.isLoadingHackathonList && <div>Loading...</div>}
+          </Group>
+        </Card.Section>
+        <Card.Section className={classes.section}>
+          <ol>{allHackathons}</ol>
```
verzano, line 95: This is one I'm sure you have thought of, but would be good to just have this whole list be it's own component.  Same thing about the other lists of course

hms-app/src/components/CategoryDetails.tsx
```diff
@@ -0,0 +1,142 @@
...
+              </p>
+            </Modal>
+            <Group position="left" mt="xl">
+              <Button color={'red'} onClick={() => setOpened(true)}>
+                Delete
+              </Button>
+              <Button>Edit</Button>
+              <Button color={'green'} onClick={() => loadSelectedCategory()}>
+                Reload
+              </Button>
```
verzano, line 135: since we have these `Delete`, `Reload`, and `Edit` buttons in a lot of places it would be good to extract them to a single component.  That way if we want to change the color or text we need only do it in one place

hms-app/src/components/NewHackathonDetails.tsx
```diff
@@ -0,0 +1,472 @@
...
+    ) {
+      setRelevantIdeaList((relevantIdeaList) => {
+        return [...relevantIdeaList, ideaData]
+      })
+    }
+  }, [ideaData])
+
+  const participantsPreviewData = hackathonData.participants?.map(
+    (participant, index) => (
+      <SimpleGrid cols={2} breakpoints={[{ maxWidth: 'sm', cols: 1 }]}>
```
verzano, line 202: same thing about the `<li>` and using `<div>`s

hms-app/src/components/AllHackathonList.tsx
LetsPoke: whole component will be deleted

hms-app/src/components/AllHackathonList.tsx
LetsPoke: done this in the new component

hms-app/src/components/AllHackathonList.tsx
LetsPoke: whole component gets deleted

verzano (approved):

PR #306, opened 2022-04-28
refactoring hackathon details

### LINKED ISSUE #302: refactoring the admin page

LetsPoke:
a lot of refactoring
more to come in future

closes #302


hms-app/src/actions/ParticipantActions.tsx
```diff
@@ -37,3 +37,20 @@ export const createParticipant = (userId: string, hackathonId: string) => {
     })
     .catch((err) => console.log(err))
 }
+
+export const getParticipantDetails = (participantID: string) => {
+  return fetch(
+    `http://localhost:4566/restapis/${api_id}/local/_user_request_/category/${participantID}`,
```
verzano, line 43: so this `http://localhost:4566/restapis/${api_id}/local/_user_request_` is repeated a lot.  You could probably extract it as a a const named something like hmsApiBaseUrl and put it where the thing that you get the `api_id` is.  Just cuts down on duplication and shortens these lines

hms-app/.env
```diff
@@ -1 +1 @@
-REACT_APP_API_ID=mxpub1nyld
\ No newline at end of file
+REACT_APP_CORE_URL=http://localhost:4566/restapis/${api_id}/local/_user_request_
```
verzano, line 1: well you would still need to replace ${api_id} with something, and I meant it a bit differently but I didn't make it to clear.  However this should work perfectly until we get to the point where we have a production and local environment.  Then we would have either just environment variables or a function that will get the correct base url :)

hms-app/src/actions/HackathonActions.tsx
```diff
@@ -1,33 +1,27 @@
-const api_id = process.env.REACT_APP_API_ID
+const core_url = process.env.REACT_APP_CORE_URL
 
 export const getListOfHackathons = () => {
```
verzano, line 3: so I think what could really make your code a bit simpler is to set the return type of these.  To make it work you would need to do something along the lines of....
- Change the signature to return a `Promise<Hackathon[]>` or something like that
- Remove the `catch` statement, this line makes it also possible for the function to return `Promise<void>`
  - You could also keep the line and rethrow the Error that you catch after you log it.  Either way if you keep the line and rethrow or remove it, the callers of this function need to handle the error somehow.
  - Another solution would be leave the line and change the signature to something like `Promise<void | Hackathon[]>` but then you need to be sure you handle the case in the frontend when this returns void in the things hat use this function
- Add another `.then()` line that converts the json to a Hackathon[].  This will add a good bit of code of course but will remove duplication from the callers of this method

hms-app/src/common/types.tsx
```diff
@@ -42,27 +42,22 @@ export type CategoryPreview = {
...
   user: User
 }
 
-export type ParticipantPreview = {
+export type Participant = {
+  userId: string
+  hackathonId: string
   id: string
-  user: User
+  creationDate: string
```
verzano, line 60: this could be a Date object, you can pass a string into `new Date()` and so long as we don't return bad stuff from the backend it should always work

hms-app/src/components/IdeaCardFoldable.tsx
```diff
@@ -126,69 +186,58 @@ export default function IdeaCardFoldable(props: IProps) {
...
+                  <Modal
+                    centered
+                    opened={opened}
+                    onClose={() => setOpened(false)}
+                    withCloseButton={false}
+                  >
+                    Are you sure you want to delete this idea?
+                    <h4>Title: {idea.title}</h4>
+                    <Button color={'red'} onClick={() => deleteSelectedIdea()}>
+                      Yes delete this idea
```
verzano, line 226:
```suggestion
                      Yes, delete this idea
```

hms-app/src/components/hackathon-details/CategoryDetails.tsx
```diff
@@ -45,21 +48,17 @@ export default function CategoryDetails(props: IProps) {
...
-          isLoadingCategoryData: false,
         })
       },
       () => {
-        setCategoryData({
-          ...categoryData,
-          errorCategoryData: true,
-          isLoadingCategoryData: false,
-        })
+        setIsError(true)
```
verzano, line 61: probably also `setIsLoading(false)`

hms-app/src/components/hackathon-details/CategoryDetails.tsx
```diff
@@ -72,27 +71,24 @@ export default function CategoryDetails(props: IProps) {
...
       )}
-      {categoryData.isLoadingCategoryData && (
+      {isLoading && (
         <div>
           <h3>Category details are loading...</h3>
         </div>
       )}
 
-      {!categoryData.isLoadingCategoryData && (
+      {!isLoading && (
```
verzano, line 91: Maybe also :thinking: 
```suggestion
      {!isLoading && !isError (
```

hms-app/src/components/hackathon-details/NewCategory.tsx
```diff
@@ -30,30 +26,22 @@ const useStyles = createStyles((theme) => ({
...
   const { classes } = useStyles()
+  const { hackathonID } = props
+  const [hasTitle, setHasTitle] = useState(false)
   const [categoryText, setCategoryText] = useState({
     title: '',
     description: '',
-    hackathonID: '',
-    titleSet: false,
-    hackathonIDSet: false,
+    hackathonID: hackathonID.toString(),
```
verzano, line 36:
```suggestion
    hackathonID: hackathonID,
```

hms-app/.env
LetsPoke: omg i replaced this after the commit -.-

hms-app/src/components/hackathon-details/NewCategory.tsx
```diff
@@ -30,30 +26,22 @@ const useStyles = createStyles((theme) => ({
   },
 }))
 
-function NewCategory() {
+function NewCategory(props: IProps) {
   const { classes } = useStyles()
+  const { hackathonID } = props
+  const [hasTitle, setHasTitle] = useState(false)
   const [categoryText, setCategoryText] = useState({
```
verzano, line 33: just `category` might be a better name :thinking:

hms-app/src/components/hackathon-details/NewCategory.tsx
```diff
@@ -127,7 +96,7 @@ function NewCategory() {
           />
         </Card.Section>
         <Group position="right" mt="xl">
-          <Button disabled={!categoryText.titleSet} onClick={submitForm}>
+          <Button disabled={!hasTitle} onClick={submitForm}>
```
verzano, line 99: so this bit I would advise to write a function that looks at the current state and determines if this should be disabled. You could do away with the `hasTitle` state hook and just do something like:

```
function submitIsEnabled(): boolean {
  return !!categoryText.title
}
```

hms-app/src/components/hackathon-details/NewHackathonDetails.tsx
```diff
@@ -50,17 +51,7 @@ const useStyles = createStyles((theme) => ({
...
-  })
-  const [deleteIDs, setDeleteIDs] = useState({
-    participantID: '',
-    categoryID: '',
-    ideaID: '',
-  })
+  const [opened, setOpened] = useState(false)
   const [hackathonData, setHackathonData] = useState({
     errorHackathonData: false,
     isLoadingHackathonData: true,
```
verzano, line 57: here I would suggest the same thing about making these their own state hooks

hms-app/src/components/hackathon-details/NewHackathonDetails.tsx
```diff
@@ -50,17 +51,7 @@ const useStyles = createStyles((theme) => ({
...
-    participantsModal: false,
-    categoriesModal: false,
-    ideasModal: false,
-  })
-  const [deleteIDs, setDeleteIDs] = useState({
-    participantID: '',
-    categoryID: '',
-    ideaID: '',
-  })
+  const [opened, setOpened] = useState(false)
```
verzano, line 54: it's not so obvious what `opened` means in this class.  I think it's referring to a modal window being displayed? maybe a more descriptive name would be helpful `modalOpened` or something along those lines

hms-app/src/components/hackathon-details/ParticipantDetails.tsx
```diff
@@ -0,0 +1,92 @@
+import React, { useState } from 'react'
+import { User } from '../../common/types'
+import { Button, Card, createStyles, Group, Modal, Text } from '@mantine/core'
+import { deleteParticipant } from '../../actions/ParticipantActions'
+
+type IProps = {
```
verzano, line 6: so the downside of defining these here and keeping them internal (and giving it the same name as all the other IProps) is that you cannot within the code validate that you are passing in correct values.  I think TS will maybe figure it out for you, but it could be done by the programmer a bit easier if these were exported as something like `ParticipantDetailsProps` and then they would have to cast their input to this component to it.  It should help catch errors sooner

PR #318, opened 2022-05-02
313 admin page

### LINKED ISSUE #313: Admin Page
Delete Hackathon
Date from:Sat Apr 30 2022 --> Start Date: Sat Apr 30 2022
to:Wed Aug 31 2022 --> End Date: Wed Aug 31 2022

User List (doesn't exist)
User Details

hackathon details editable

hackathon create date picker doesnt check for month

LetsPoke:
- added user list
- fixed month display of datepicker
- make everything editable

closes #313


hms-app/src/actions/CategoryActions.tsx
```diff
@@ -1,9 +1,9 @@
 const core_url = process.env.REACT_APP_CORE_URL
 
 export const addCategory = (props: {
+  contextID: string
```
verzano, line 4: I think still naming it hackathonId is good, then you can be sure without other information of what value to pass into here

hms-app/src/actions/HackathonActions.tsx
```diff
@@ -28,21 +28,45 @@ export const getHackathonDetails = (hackathonID: string) => {
     .catch((err) => console.log(err))
 }
 
-export const createHackathon = (props: {
-  endDate: string
-  title: string
-  startDate: string
-}) => {
+export const createHackathon = (
+  hackathonText: string,
```
verzano, line 32: hackathonText is a bit vague, hackathonTitle or just title is a bit more descriptive

hms-app/src/actions/IdeaActions.tsx
```diff
@@ -64,3 +64,37 @@ export const createIdea = (
...
+      Accept: 'application/json',
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      title: idea.title,
+      description: idea.description,
+      problem: idea.problem,
+      goal: idea.goal,
+      requiredSkills: skills,
+      categoryId: categories.toString(),
```
verzano, line 93: this will do something weird, as it is still a single category in the backend

hms-app/src/actions/IdeaActions.tsx
```diff
@@ -64,3 +64,37 @@ export const createIdea = (
...
+  idea: {
+    ownerId: string
+    hackathonId: string
+    title: string
+    description: string
+    problem: string
+    goal: string
+  },
+  skills: string[],
+  categories: string[]
```
verzano, line 79: why are they separate? are required skill and the category a part of the idea? as well as the id itself too?

hms-app/src/actions/CategoryActions.tsx
LetsPoke: the problem is that conextID sometimes is a category id and sometimes a hackathon id and have to name both the same

hms-app/src/actions/IdeaActions.tsx
LetsPoke: because of the checkbox input type it would need an extra step to put them all in the idea object

hms-app/src/actions/IdeaActions.tsx
LetsPoke: lets assume there is only 1 category in the frontend for now

hms-app/src/actions/CategoryActions.tsx
verzano: but for this function you only ever use it as a hackathonID

hms-app/src/actions/IdeaActions.tsx
verzano: yeah, but this will convert it to a string like `["aabeb-12234nb..."]` with the brackets around it and the backend won't (or shouldn't) like it

hms-app/src/actions/IdeaActions.tsx
verzano: I would honestly argue that is ok.  Or I would say to not allow you to pass in an incomplete Idea.  When you make it like this you have added another concept of what an `idea` is.  You could instead call this `ideaStringData` or something but honestly thats not great either

hms-app/src/actions/CategoryActions.tsx
```diff
@@ -64,3 +64,25 @@ export const getCategoryDetails = (categoryID: string) => {
     })
     .catch((err) => console.log(err))
 }
+
+export const editCategory = (props: {
+  contextID: string
```
verzano, line 69: and here it is for sure a categoryID always so you could name it that

hms-app/src/actions/HackathonActions.tsx
```diff
@@ -28,21 +28,45 @@ export const getHackathonDetails = (hackathonID: string) => {
...
+  })
+    .then((response) => {
+      return response.json()
+    })
+    .catch((err) => console.log(err))
+}
+
+export const editHackathon = (
+  hackathonID: string,
+  hackathonText: string,
```
verzano, line 56: same thing here about `hackathonText`, `hackathonTitle` or just `title`.

hms-app/src/actions/UserActions.tsx
```diff
@@ -0,0 +1,75 @@
...
+export const editUser = (
+  contextID: string,
+  user: {
+    lastName: string
+    firstName: string
+    emailAddress: string
+    roles: any[]
+    imageUrl: string
+  },
+  skills: string[]
```
verzano, line 54: so a similar thing here, contextID is a bit misleading and you won't know just by looking at it what value should go there.  So userID would be better.  In addition this could all just be one user object, or all individual fields instead of the mix

hms-app/src/components/admin-tools/UserForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+      (data) => {
+        setIsError(false)
+        setIsLoading(false)
+        setUser({
+          lastName: data.lastName,
+          firstName: data.firstName,
+          emailAddress: data.emailAddress,
+          roles: data.roles,
+          imageUrl: data.imageUrl,
+        })
```
verzano, line 67: shouldn't you also set the skills in here?

verzano: like the initial values of them should be set

hms-app/src/components/admin-tools/UserForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+  useEffect(() => {
+    loadSelectedUser()
+    loadAvailableSkills()
+    setIsLoading(true)
+  }, [])
+
+  function handleChange(event: React.ChangeEvent<HTMLTextAreaElement>) {
+    setUser((prevUserText) => ({
+      ...prevUserText,
+      [event.target.name]: event.target.value,
```
verzano, line 85:
```suggestion
    setUser((prevUser) => ({
      ...prevUser,
      [event.target.name]: event.target.value,
```

hms-app/src/components/admin-tools/UserForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+} from '@mantine/core'
+import React, { useEffect, useState } from 'react'
+import { showNotification, updateNotification } from '@mantine/notifications'
+import { CheckIcon } from '@modulz/radix-icons'
+import { editUser, getUserDetails } from '../../actions/UserActions'
+import { SkillPreview } from '../../common/types'
+import { getListOfSkills } from '../../actions/SkillActions'
+
+type IProps = {
+  contextID: string
```
verzano, line 18: should name this userID

hms-app/src/components/admin-tools/UserForm.tsx
```diff
@@ -0,0 +1,197 @@
+import {
```
verzano, line 1: this might be better named as `EditUserForm` since it's purpose is specifically to do that

verzano: it might also make sense located outside the `admin-tools` folder as likely we would allow users to edit their own information which they won't need to be an Admin to do :thinking:

hms-app/src/components/admin-tools/UserForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+  function submitIsEnabled(): boolean {
+    return !!user.firstName
+  }
+
+  const loadAvailableSkills = () => {
+    getListOfSkills().then((data) => {
+      setAvailableSkills({
+        ...availableSkills,
+        skills: data.skills,
+      })
```
verzano, line 122: I expect this won't work quite right.  availableSkills is meant to be an array not an object, but this will create an object with a field inside it called `skills` that is an string[] and another string[] with no name... all you should do really is `setAvailableSkills(data.skills)`

verzano: we should assume in this instance that the data we get from the backend is the complete list of skills so we can just wipe out whatever data is currently stored in `availableSkills` and replace it with the data from the backend

hms-app/src/components/admin-tools/UserForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+    })
+  }
+
+  const skillsList = availableSkills.skills.map((skill, index) => [
```
verzano, line 126: this should then just be:
```suggestion
  const skillsList = availableSkills.map((skill, index) => [
```

verzano: also I'm pretty sure that when you write these if you don't need the `index` you can just leave it out :thinking:

hms-app/src/components/admin-tools/UserForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+          </Card.Section>
+          <Card.Section className={classes.section}>
+            <Textarea
+              label="Last Name"
+              mt="sm"
+              placeholder="Last Name"
+              maxRows={1}
+              autosize
+              onChange={handleChange}
+              name="description"
```
verzano, line 160:
```suggestion
              name="lastName"
```

hms-app/src/components/admin-tools/UserForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+          <Card.Section className={classes.section}>
+            <Textarea
+              label="First Name"
+              mt="sm"
+              required
+              placeholder="First Name"
+              maxRows={1}
+              autosize
+              onChange={handleChange}
+              name="title"
```
verzano, line 148:
```suggestion
              name="firstName"
```

hms-app/src/components/admin-tools/UserForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+          </Card.Section>
+          <Card.Section className={classes.section}>
+            <Textarea
+              label="E-mail"
+              mt="sm"
+              placeholder="E-mail"
+              maxRows={1}
+              autosize
+              onChange={handleChange}
+              name="description"
```
verzano, line 172:
```suggestion
              name="emailAddress"
```

hms-app/src/components/admin-tools/UserForm.tsx
```diff
@@ -0,0 +1,197 @@
...
+              label="Skills"
+              spacing="md"
+              onChange={setSkills}
+              required
+            >
+              {skillsList}
+            </CheckboxGroup>
+          </Card.Section>
+          <Group position="right" mt="xl">
+            <Button disabled={!submitIsEnabled()} onClick={editThisUser}>
```
verzano, line 189: the one thing about all this though is that I still haven't implemented an editUser endpoint :P

hms-app/src/components/admin-tools/UserDetails.tsx
```diff
@@ -0,0 +1,220 @@
...
+        setUser({
+          id: data.id,
+          lastName: data.lastName,
+          firstName: data.firstName,
+          emailAddress: data.emailAddress,
+          roles: data.roles,
+          skills: data.skills,
+          imageUrl: data.imageUrl,
+          creationDate: data.creationDate,
+        })
```
verzano, line 71:
```suggestion
        setUser(data)
```

verzano: it might complain about this though.  To fix it you would need to make it so that `getUserDetails()` returns `Promise<User>`

hms-app/src/components/admin-tools/UserDetails.tsx
```diff
@@ -0,0 +1,220 @@
...
+      Are you sure you want to delete this user?
+      <h4>
+        Name: {user.firstName} {user.lastName}
+      </h4>
+      <h4>E-mail: {user.emailAddress}</h4>
+      <Button color={'red'} onClick={() => deleteSelectedUser()}>
+        Yes delete this user
+      </Button>
+      <p>
+        (This window will automatically closed as soon as the hackathon is
```
verzano, line 109:
```suggestion
        (This window will automatically closed as soon as the user is
```

hms-app/src/components/admin-tools/UserDetails.tsx
```diff
@@ -0,0 +1,220 @@
...
+      opened={editModalOpened}
+      onClose={() => setEditModalOpened(false)}
+      withCloseButton={false}
+    >
+      Edit User
+      <UserForm contextID={userID} />
+      {isUserLoading && <div>Loading...</div>}
+      <p>
+        (This window will automatically close as soon as the category is
+        deleted)
```
verzano, line 127:
```suggestion
        (This window will automatically close as soon as the user is
        updated)
```

hms-app/src/components/admin-tools/UserDetails.tsx
```diff
@@ -0,0 +1,220 @@
...
+      </p>
+    </Modal>
+  )
+
+  return (
+    <>
+      {isUserError && (
+        <div>
+          <h3>Error loading hackathons</h3>
```
verzano, line 136:
```suggestion
          <h3>Error loading user</h3>
```

hms-app/src/components/admin-tools/UserDetails.tsx
```diff
@@ -0,0 +1,220 @@
...
+      <p>
+        (This window will automatically close as soon as the category is
+        deleted)
+      </p>
+    </Modal>
+  )
+
+  return (
+    <>
+      {isUserError && (
```
verzano, line 134: this maybe should be `isUserError && !isUserLoading && (` then it will be hidden as soon as you try to reload :thinking: the same thing would apply to all the other components where we have `isLoading` and `isError`

hms-app/src/components/admin-tools/UserDetails.tsx
```diff
@@ -0,0 +1,220 @@
...
+          <Card.Section className={classes.section}>
+            <Text mt="md" className={classes.label} color="dimmed">
+              Name
+            </Text>
+            <Text size="md" mt="xs">
+              {user.firstName} {user.lastName}
+            </Text>
+            <Text size={'xs'}>ID: {user.id}</Text>
+            <Text size={'xs'}>
+              Creation Date: {user.creationDate?.toString()}
```
verzano, line 157: there is a ticket about figuring out which fields can be null.  Once we have that decided we can remove some of the `?` uses

hms-app/src/components/admin-tools/UserDetails.tsx
```diff
@@ -0,0 +1,220 @@
...
+
+          <Card.Section className={classes.section}>
+            <Group position="left" mt="xl">
+              {deleteModal}
+              <Button color={'red'} onClick={() => setDeleteModalOpened(true)}>
+                Delete
+              </Button>
+              {editModal}
+              <Button onClick={() => setEditModalOpened(true)}>Edit</Button>
+              {!isUserLoading && (
```
verzano, line 208: this will always evaluate to `true` here because it is within the other `{!isUserLoading` check

hms-app/src/components/admin-tools/HackathonForm.tsx
```diff
@@ -30,28 +30,26 @@ const useStyles = createStyles((theme) => ({
...
-function NewHackathon() {
+type IProps = { context: string; hackathonID: string | null }
+
+function HackathonForm(props: IProps) {
+  const { context, hackathonID } = props
   const { classes } = useStyles()
   const today = new Date()
   const [startDateValue, setStartDateValue] = useState<Date | null>(new Date())
   const [endDateValue, setEndDateValue] = useState<Date | null>(new Date())
   const [hackathonText, setHackathonText] = useState({
```
verzano, line 41: `hackathon` would be a better name, or draftHackathon possibly since the intention is to edit or create a hackathon

verzano: ah, but now that I see how you modified this... you could just change this to `title` and do `useState('')` or something instead of making this an object

hms-app/src/components/admin-tools/HackathonForm.tsx
```diff
@@ -30,28 +30,26 @@ const useStyles = createStyles((theme) => ({
   },
 }))
 
-function NewHackathon() {
+type IProps = { context: string; hackathonID: string | null }
+
+function HackathonForm(props: IProps) {
```
verzano, line 35: Might be able to make a better name for this as well, but it performs two functions to both create and edit hackathons.  I guess you are using `Form` in the sense of a `Form` on a website that allows data entry... maybe thats ok then :thinking:

hms-app/src/pages/IdeaPortal.tsx
```diff
@@ -57,7 +57,7 @@ function IdeaPortal() {
       (data) => {
         setHackathonList({
           ...hackathonList,
-          hackathons: data.hackathons,
+          users: data.hackathons,
```
verzano, line 60: pretty sure these aren't users

hms-app/src/pages/IdeaPortal.tsx
```diff
@@ -161,12 +161,10 @@ function IdeaPortal() {
     return item.title?.toLowerCase().includes(searchTerm.toLowerCase())
   })
 
-  const data = hackathonList.hackathons.map(
-    (hackathon, index) => hackathon.title
-  )
+  const data = hackathonList.users.map((hackathon, index) => hackathon.title)
```
verzano, line 164: `data` isn't a great name for this.  Maybe `hackathonTitles`? but doing this also feels not great because you lose the linking between ID and title :thinking:

hms-app/src/pages/IdeaPortal.tsx
```diff
@@ -161,12 +161,10 @@ function IdeaPortal() {
...
 
   const selectChange = (value: string) => {
-    const getHackathon = hackathonList.hackathons.filter((hackathon) => {
+    const getHackathon = hackathonList.users.filter((hackathon) => {
```
verzano, line 167: ok so I don't know exactly how to do it for these components, but normally in a dropdown list you can give the list an object and then tell it how to display the object.  Doing this means that when you select an item in the dropdown you get the object passed into the callback instead of just a string.  Then you don't need to search for the object like you are doing here

hms-app/src/pages/YourIdeas.tsx
```diff
@@ -46,6 +45,10 @@ function YourIdeas() {
   } as Idea)
   const [relevantIdeaList, setRelevantIdeaList] = useState([] as Idea[])
 
+  console.log('selected Hackweek', selectedHackweek)
+  console.log('hackathonList', hackathonList)
+  console.log('hackathonData', hackathonData)
```
verzano, line 50: ewwwww

hms-app/src/pages/YourIdeas.tsx
```diff
@@ -217,11 +224,7 @@ function YourIdeas() {
           <h2>Your Ideas ({filteredIdeas.length})</h2>
 
           <div>
-            <IdeaCardList
-              ideas={filteredIdeas}
-              columnSize={6}
-              type={'idea-portal'}
-            />
+            <IdeaCardList ideas={filteredIdeas} columnSize={6} type={'owner'} />
```
verzano, line 227: this `type` should also be an enum to make using it easier

hms-app/src/components/admin-tools/UserForm.tsx
LetsPoke: :O

verzano (approved):

PR #330, opened 2022-05-09
324 add a presentation tool

### LINKED ISSUE #324: add a presentation tool
display ideas as presentation

- add a section in the admin page
- select a hackathon and get directed to a page with ideas that belongs to the hackathon
- add a fullscreen toogle
- see the ideas in fullscreen in a presentation style
- add buttons/whatever to navigate through the ideas

LetsPoke:
in adminpage in hackathon list in hackathon details in Idea List you now can go to a fullscreen page (after all ideas loaded)
- added a new page
- added a idea slider
- added a fullscreen toggle

closes #324


hms-app/src/components/card- details/HackathonDetails.tsx
```diff
@@ -361,6 +362,19 @@ export default function HackathonDetails(props: IProps) {
                   </Text>
                 }
               >
+                <Button
+                  mb={20}
+                  onClick={() =>
+                    localStorage.setItem(
```
verzano, line 368: nice!

verzano: however I would say to push this somewhere else.  Likely it should only ever be done by the `actions` portion of the frontend.  That should reduce a bit of code duplication that could pop up if multiple places can load ideas from the backend.  In fact it should probably happen something like this

1. ask some action function for ideas
2. check if they are in localstorage
  a. if they are use what is in localstorage and don't call the backend
  b. if not then call the backend to get the ideas
    I. store the ideas in localstorage
3. return the ideas to the caller

then we should have a boolean that can be passed in (set to false if nothing is passed in) that can force reloading of ideas (so it just skips checking in localstorage).

this could be done for every loadXxx() function

hms-app/src/pages/Presentations.tsx
```diff
@@ -0,0 +1,203 @@
...
+                  <Text
+                    mt="sm"
+                    style={{
+                      backgroundColor: 'white',
+                      height: '15vh',
+                    }}
+                  >
+                    <ul>
+                      {idea.requiredSkills?.map((skill, index) => (
+                        <li>{skill.name}</li>
```
verzano, line 133:
```suggestion
                        <li key={skill.id}>{skill.name}</li>
```

hms-app/src/pages/Presentations.tsx
```diff
@@ -0,0 +1,203 @@
...
+                            <Avatar
+                              color="indigo"
+                              radius="xl"
+                              size="md"
+                              src={
+                                'https://avatars.githubusercontent.com/u/10353856?s=460&u=88394dfd67727327c1f7670a1764dc38a8a24831&v=4'
+                              }
+                            />
+                            {participant.user.firstName}{' '}
+                            {participant.user.lastName}
```
verzano, line 170: maybe this should get converted into a function as part of a `user` class (or some function that accepts a user as input).  Then it would do something a bit like:

```javascript
function renderName(user: User): string {
  user.firstName + (user.lastName ? ' ' + user.lastName : '');
}
```

verzano: this just prevents an extra ` ` (space) from being included when the lastName is null/undefined

hms-app/src/pages/Presentations.tsx
```diff
@@ -0,0 +1,203 @@
...
+  },
+}))
+
+export default function Presentations() {
+  const { classes } = useStyles()
+  const { ref, toggle, fullscreen } = useFullscreen()
+  // @ts-ignore
+  const allIdeas = JSON.parse(localStorage.getItem('ideas'))
+
+  function getIdeasMap() {
```
verzano, line 41: more of an `ideas list` rather than a `map`

Issue #341, opened 2022-05-12
filter hackathon select in your ideas


LetsPoke:
user should only have the option to select one of the hackathons in which he is listed as participant (in the your ideas page)

and the user should only be able to create new ideas for actual/upcoming hackathons not for past ones

Issue #260, opened 2022-04-20
filter "your ideas"


LetsPoke:
in the your ideas page only display the ideas that connects to the specific user

PR #354, opened 2022-05-17
311 hackathon header

### LINKED ISSUE #311: hackathon header
Current Hackathon is always loaded even when nothing selected
Single Hackathon dropdown in nav bar at the top
Should display only current and future hackathons sorted by start date
add `previous hackathons` and `future hackathons` endpoints (or add this to the request body)

LetsPoke:
a lot of stuff

closes #311 #260 #341


hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -27,38 +31,53 @@ export default function HackathonSelectDropdown(setHackathonID: Props) {
     )
   }
 
-  const hackathonMap = hackathonList.map((hackathon, index) => ({
+  const hackathonDateCurrentAndFuture = hackathonList.filter((hackathon) => {
+    return new Date(hackathon.endDate) > today
+  })
+
+  const hackathonDatePast = hackathonList.filter((hackathon) => {
```
KBernaszuk, line 38: Same as above but pastHacaktons

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -27,38 +31,53 @@ export default function HackathonSelectDropdown(setHackathonID: Props) {
...
+    return new Date(hackathon.endDate) < today
+  })
+
+  const hackathonMap = (
+    context === 'archive'
+      ? hackathonDatePast
+      : context === 'idea-portal'
+      ? hackathonDateCurrentAndFuture
+      : hackathonList
```
KBernaszuk, line 47: This is hard to read, wdyt about:
```suggestion
const hackathonMap = getHackatonsForContext()
```

```
function getHackatonsForContext(): HackathonPreview[] {
    switch(context){
      case 'archive':
        return hackathonDatePast;
      case 'ida-portal':
        return futureAndPresentHackatons;
    }
    return hackathonList;
  }
```

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -27,38 +31,53 @@ export default function HackathonSelectDropdown(setHackathonID: Props) {
...
+        month: '2-digit',
+        day: '2-digit',
+      }) +
       '-' +
-      new Date(hackathon.endDate).toLocaleDateString(),
+      new Date(hackathon.endDate).toLocaleString(undefined, {
+        year: '2-digit',
+        month: '2-digit',
+        day: '2-digit',
+      }),
```
KBernaszuk, line 63: Let's extract this so it's more readable and the result type is clear
```
const hackathonMap = getHackatonsForContext()
    .map(hackathon => mapHackatonToSelectItem(hackathon))
```
example impl:
```
function mapHackatonToSelectItem(hackathon: HackathonPreview): SelectItem {
    return {
      value: hackathon.id,
      label: hackathon.title +
        ' ' +
        new Date(hackathon.startDate).toLocaleDateString() +
        '-' +
        new Date(hackathon.endDate).toLocaleDateString()
    }
  }
```
Note you can use `toLocaleDateString()` to return just the date

hms-app/src/components/RelevantIdeasLoader.tsx
```diff
@@ -4,30 +4,30 @@ import { Hackathon, Idea } from '../common/types'
 import { getIdeaDetails } from '../actions/IdeaActions'
 
 type Props = {
-  selectedHackweek: string
+  selectedHackathonID: string
```
KBernaszuk, line 7: If we generally follow camelcase, we should change to it all the fields with `Id` in the suffix
```suggestion
  selectedHackathonId: string
```

hms-app/src/components/RelevantIdeasLoader.tsx
```diff
@@ -39,15 +39,15 @@ export default function RelevantIdeasLoader({
...
       setHackathonData({
         id: data.id,
         title: data.title,
-        startDate: data.startDate,
-        endDate: data.endDate,
+        startDate: new Date(data.startDate),
+        endDate: new Date(data.endDate),
         participants: data.participants,
         categories: data.categories,
         ideas: data.ideas,
```
KBernaszuk, line 53: You can make `getHackatonDetails()` return a `Promise<Hackathon>` and then you can have here simply:
```
  const hackathonMap = getHackatonsForContext()
    .map(hackathon => mapHackatonToSelectItem(hackathon))
```
So you don't need to copy one object state to other by hand

hms-app/src/components/RelevantIdeasLoader.tsx
```diff
@@ -4,30 +4,30 @@ import { Hackathon, Idea } from '../common/types'
...
+  selectedHackathonID,
   setRelevantIdea,
   setHackathon,
   setLoading,
 }: Props) {
-  const [relevantIdeaList, setRelevantIdeaList] = useState([] as Idea[])
+  const [relevantIdeaList, setRelevantIdeaList] = useState<Idea[]>([])
   const [isLoading, setIsLoading] = useState(true)
-  const [hackathonData, setHackathonData] = useState({
+  const [hackathonData, setHackathonData] = useState<Hackathon>({
```
KBernaszuk, line 21: You can use custom hook `useLocalStorage("key", defaultValue);` to skip manual checks for the presence of a value :D

hms-app/src/components/RelevantIdeasLoader.tsx
```diff
@@ -77,14 +77,17 @@ export default function RelevantIdeasLoader({
   }
 
   useEffect(() => {
-    loadSelectedHackathonData()
-  }, [selectedHackweek])
+    localStorage.getItem(selectedHackathonID)
```
KBernaszuk: You won't need this if you use the custom hook, as it will be handled by default

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -27,38 +31,63 @@ export default function HackathonSelectDropdown(setHackathonID: Props) {
...
-    value: hackathon.id,
-    label:
-      hackathon.title +
-      ' ' +
-      new Date(hackathon.startDate).toLocaleDateString() +
-      '-' +
-      new Date(hackathon.endDate).toLocaleDateString(),
-  }))
+  const futureAndPresentHackathons = hackathonList.filter((hackathon) => {
+    return new Date(hackathon.endDate) > today
```
verzano, line 35: you probable want >= today and maybe even want to force today to be the start of the day

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -27,38 +31,63 @@ export default function HackathonSelectDropdown(setHackathonID: Props) {
...
+  const pastHackathons = hackathonList.filter((hackathon) => {
+    return new Date(hackathon.endDate) < today
+  })
+
+  function getHackatonsForContext(): HackathonPreview[] {
+    switch (context) {
+      case 'archive':
+        return pastHackathons
+      case 'idea-portal':
+        return futureAndPresentHackathons
```
verzano, line 47: these should probably be enum values and not magic strings

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -27,38 +31,63 @@ export default function HackathonSelectDropdown(setHackathonID: Props) {
...
+      label:
+        hackathon.title +
+        ' ' +
+        new Date(hackathon.startDate).toLocaleString(undefined, {
+          year: 'numeric',
+          month: '2-digit',
+          day: '2-digit',
+        }) +
+        '-' +
+        new Date(hackathon.endDate).toLocaleString(undefined, {
```
verzano, line 64:
```suggestion
        hackathon.startDate.toLocaleString(undefined, {
          year: 'numeric',
          month: '2-digit',
          day: '2-digit',
        }) +
        '-' +
        ackathon.endDate.toLocaleString(undefined, {
```

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -27,38 +31,63 @@ export default function HackathonSelectDropdown(setHackathonID: Props) {
...
-      '-' +
-      new Date(hackathon.endDate).toLocaleDateString(),
-  }))
+  const futureAndPresentHackathons = hackathonList.filter((hackathon) => {
+    return new Date(hackathon.endDate) > today
+  })
 
-  const selectChange = (value: string) => {
-    setSelectedHackweek(value)
+  const pastHackathons = hackathonList.filter((hackathon) => {
```
verzano, line 38: so you shouldn't create both of these variables.  At best you could lazy load into variables, but likely this part you are doing should be done within the `getHackathonsForContext()` funciton

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -27,38 +31,63 @@ export default function HackathonSelectDropdown(setHackathonID: Props) {
...
+    return new Date(hackathon.endDate) < today
+  })
+
+  function getHackatonsForContext(): HackathonPreview[] {
```
verzano, line 42:
```suggestion
  function getHackathonsForContext(): HackathonPreview[] {
```

hms-app/src/components/HackathonSelectDropdown.tsx
verzano: the `archive` and `idea-portal` I mean

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -27,38 +31,63 @@ export default function HackathonSelectDropdown(setHackathonID: Props) {
...
+        '-' +
+        new Date(hackathon.endDate).toLocaleString(undefined, {
+          year: 'numeric',
+          month: '2-digit',
+          day: '2-digit',
+        }),
+    }
   }
 
+  const hackathonMap = getHackatonsForContext().map((hackathon) =>
```
verzano, line 72: would you ever call `getHackathonsForContext()` and not immediately call this map function on it? if not then you could probably just have `getHackathonsForContext()` also do the map

hms-app/src/components/HackathonSelectDropdown.tsx
KBernaszuk: Oh.. nice catch 🚀  When I went trough it earlier, i tough these were functions accepting `hackatonList` as parm. In this case you need to either make them functions or use them directly as expressions in `getHackathonsForContext()` directly.

hms-app/src/components/RelevantIdeasLoader.tsx
LetsPoke: cant make this work sry

hms-app/src/components/HackathonSelectDropdown.tsx
LetsPoke: no

verzano: endDate and startDate are already `Date` objects so you shouldn't need to convert them into another Date object

verzano: the problem might be that when you do the fetch for the data you don't convert them to Date objects immediately.  The type defines them as being Date objects so you should always assume they are, you just have to ensure that when you create an instance of that type you set the values properly

LetsPoke: the problem is not the type as i understand
the problem is when we want to display them with DD.MM.YYYY this is the way we have to do it 
otherwise we would have a different format

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -1,17 +1,26 @@
 import React, { useEffect, useState } from 'react'
 import { getListOfHackathons } from '../actions/HackathonActions'
-import { Select } from '@mantine/core'
+import { Select, SelectItem } from '@mantine/core'
 import { HackathonPreview } from '../common/types'
 
 type Props = {
-  setHackathonID: (hackthonID: string) => void
+  setHackathonId: (hackthonID: string) => void
+  context: string
```
verzano, line 8: this should have the type of the enum you specified

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -1,17 +1,26 @@
...
 }
 
-export default function HackathonSelectDropdown(setHackathonID: Props) {
-  const [selectedHackweek, setSelectedHackweek] = useState('')
+enum Enum {
```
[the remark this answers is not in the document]
verzano: it should also be exported so that users of this component can pass the value in

hms-app/src/components/HackathonSelectDropdown.tsx
verzano: doing this shouldn't have any effect on the format.  That is because when you call `toLocalDateString` you force the format to be a certain way

hms-app/src/components/HeaderMenu.tsx
```diff
@@ -116,6 +116,9 @@ export default function HeaderMenu({ links }: HeaderSearchProps) {
       <Container>
         <div className={classes.inner}>
           <h1>HMS</h1>
+          <Button onClick={() => localStorage.clear()}>
```
verzano, line 119: is it for testing? you could do this through the dev console by clearing out the localstorage in the dev console

hms-app/src/components/RelevantIdeasLoader.tsx
```diff
@@ -103,9 +95,11 @@ export default function RelevantIdeasLoader({
 
   useEffect(() => {
     setRelevantIdea(relevantIdeaList)
-    setHackathon(hackathonData)
+  }, [relevantIdeaList])
+
+  useEffect(() => {
     setLoading(isLoading)
-  }, [relevantIdeaList, hackathonData, isLoading])
+  }, [isLoading])
```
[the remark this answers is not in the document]
verzano: or at least it would trigger twice every time?

hms-app/src/components/card-details/HackathonDetails.tsx
```diff
@@ -50,9 +50,15 @@ const useStyles = createStyles((theme) => ({
   },
 }))
 
+enum Enum {
```
verzano, line 53: should also have a bit better name than just `Enum`, `HackathonDetailsType` maybe

hms-app/src/components/card-details/HackathonDetails.tsx
```diff
@@ -24,7 +24,7 @@ import CategoryDetails from './CategoryDetails'
 import { Link } from 'react-router-dom'
 
 type IProps = {
-  hackathonID: string
+  hackathonId: string
   type: string
```
verzano, line 28: should be of the type of the enum

hms-app/src/components/card-details/HackathonDetails.tsx
```diff
@@ -275,19 +281,19 @@ export default function HackathonDetails(props: IProps) {
...
+              Start Date: {new Date(hackathonData.startDate).toDateString()} End
+              Date: {new Date(hackathonData.endDate).toDateString()}
             </h2>
             <h2>All Ideas ({hackathonData.ideas?.length})</h2>
 
             <IdeaCardList
               ideas={relevantIdeaList}
               columnSize={6}
-              type={'Archive'}
+              type={Enum.Archive}
```
verzano, line 296: is `archive` the right choice here? These are the header details, archive sounds wrong.  I would also think that whatever the value is should actually come from the IdeaCardList file

hms-app/src/components/card-details/UserDetails.tsx
```diff
@@ -71,15 +70,15 @@ export default function UserDetails(props: { userID: string }) {
...
   const deleteSelectedUser = () => {
-    deleteUser(userID).then((data) => {
+    deleteUser(userId).then((data) => {
       setDeleteModalOpened(false)
     })
   }
 
   useEffect(() => {
     loadSelectedUser()
     setIsUserLoading(true)
```
verzano, line 80: I know this isn't part of this PR, but this feels wrong.  By the time that `loadSelectedUser()` finishes execution the user is fully loaded.  However, right after that we set the isLoading flag for the user to true, which implies it is still loading.  Instead, the first line in the `loadSelectedUser()` function should set that value to true

hms-app/src/components/lists/AllHackathonList.tsx
```diff
@@ -21,6 +21,10 @@ const useStyles = createStyles((theme) => ({
   },
 }))
 
+enum Enum {
+  FullInfo = 'FULLINFO',
```
verzano, line 25: Maybe easier to read
```suggestion
  FullInfo = 'FULL_INFO',
```

hms-app/src/components/lists/AllHackathonList.tsx
```diff
@@ -21,6 +21,10 @@ const useStyles = createStyles((theme) => ({
   },
 }))
 
+enum Enum {
```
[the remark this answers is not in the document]
verzano: actually, this enum would make more sense inside the HackathonDetails file

hms-app/src/pages/Archive.tsx
```diff
@@ -3,22 +3,26 @@ import HackathonDetails from '../components/card-details/HackathonDetails'
...
+      <HackathonSelectDropdown
+        setHackathonId={setSelectedHackathonId}
+        context={Enum.Archive}
+      />
 
       <div>
         <HackathonDetails
-          hackathonID={selectedHackweek.toString()}
-          type={'header'}
+          hackathonId={selectedHackathonId.toString()}
```
verzano, line 24:
```suggestion
          hackathonId={selectedHackathonId}
```

hms-app/src/pages/Archive.tsx
```diff
@@ -3,22 +3,26 @@ import HackathonDetails from '../components/card-details/HackathonDetails'
...
+enum Enum {
+  Archive = 'ARCHIVE',
+  Header = 'HEADER',
+}
 
-  const setHackathonID = (hackthonID: string) => {
-    setSelectedHackweek(hackthonID)
-  }
+export default function Archive() {
+  const [selectedHackathonId, setSelectedHackathonId] = useState('')
```
verzano, line 12: likely we should set the value to something if it has no value :thinking:

hms-app/src/pages/Archive.tsx
```diff
@@ -3,22 +3,26 @@ import HackathonDetails from '../components/card-details/HackathonDetails'
 import HackathonSelectDropdown from '../components/HackathonSelectDropdown'
 import { Title } from '@mantine/core'
 
-export default function Archive() {
-  const [selectedHackweek, setSelectedHackweek] = useState('')
+enum Enum {
```
verzano, line 6: this enum should be imported from the HackathonSelectDropdown and not recreated here

hms-app/src/pages/Home.tsx
```diff
@@ -1,6 +1,66 @@
...
+    getHackathonDetails(selectedHackathonId).then((data) => {
+      setHackathonData(data)
+    })
+  }
+
+  const getNextHackathon = hackathonList.filter((hackathon) => {
+    return new Date(hackathon.startDate) > today
+  })
+
+  const timeTillNextHackathon = !isLoading
```
verzano, line 41: when it isLoading it should probably return nothing :thinking:

hms-app/src/actions/IdeaActions.tsx
```diff
@@ -10,9 +10,7 @@ export const getIdeaList = (hackathonID: string) => {
       'Content-Type': 'application/json',
     },
   })
-    .then((data) => {
-      return data.json()
-    })
+    .then((data) => data.json())
```
verzano, line 13: so in general this is the issue and the reason why you need the `new Date()` everywhere.  The JSON specs do not have a way to represent a Date object, all they have is a string representation of the date.  What you have to do is create the Date objects here.  If you don't then you end up violating the contracts you defined elsewhere when you created the different types.

it might seem tedious to do it, but if you do it here and properly create your objects you only need to do it once and there won't be any weirdness elsewhere

verzano: so instead of just returning `data.json()` you should specify a strict return type in the function declaration and then do something like (it will be different for each function):
```suggestion
    .then((data) => {
      cosnt json = data.json());
      return {
        id: json.id,
        startDate: new Date(json.startDate);
        ...
      }
    }
```

hms-app/src/components/RelevantIdeasLoader.tsx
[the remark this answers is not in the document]
LetsPoke: 2 different things, changed name to clarify

hms-app/src/pages/Archive.tsx
LetsPoke: in best case to the last loaded hackathon but still struggling with saving it

hms-app/src/pages/Home.tsx
LetsPoke: it has to return a number even when its loading

hms-app/src/actions/HackathonActions.tsx
```diff
@@ -1,31 +1,39 @@
...
       'Content-Type': 'application/json',
     },
   })
-    .then((data) => {
-      return data.json()
+    .then((data) => data.json())
+    .then((json) => parseHackathons(json.hackathons))
+    .catch((err) => {
+      console.log(err)
+      throw err
```
verzano, line 17: @KBernaszuk I said we could throw here, but I don't really know if that's the right approach in react when things fail from the backend :thinking:

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -27,38 +38,55 @@ export default function HackathonSelectDropdown(setHackathonID: Props) {
...
-      ' ' +
-      new Date(hackathon.startDate).toLocaleDateString() +
-      '-' +
-      new Date(hackathon.endDate).toLocaleDateString(),
-  }))
+  function getHackathonsSelectItems(): SelectItem[] {
+    switch (context) {
+      case HackathonDropdownMode.Archive:
+        return hackathonList
+          .filter((hackathon) => new Date(hackathon.endDate) < new Date(today))
```
verzano, line 45: now this should for sure work without the `new Date()` (around hackathon.endDate and for sure around today)

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -1,22 +1,33 @@
...
+
+export default function HackathonSelectDropdown({
+  setHackathonId,
+  context,
+}: Props) {
   const [isError, setIsError] = useState(false)
   const [isLoading, setIsLoading] = useState(true)
-  const [hackathonList, setHackathonList] = useState([] as HackathonPreview[])
+  const [hackathonList, setHackathonList] = useState<HackathonPreview[]>([])
+  const today = new Date().setHours(0, 0, 0, 0)
```
verzano, line 25: the docs say that `setHours` will return the number of millis since the epoch, so `today` won't be a date anymore.  You should do something more like...
```suggestion
  const today = new Date();
  today.setHours(0, 0, 0, 0)
```

hms-app/src/components/RelevantIdeasLoader.tsx
```diff
@@ -39,52 +39,44 @@ export default function RelevantIdeasLoader({
...
+  }, [selectedHackathonId])
+
+  useEffect(() => {
+    localStorage.getItem(hackathonData.id)
+      ? console.log(
+          'id exist',
+          JSON.parse(localStorage.getItem(hackathonData.id)!)
+        )
+      : localStorage.setItem(hackathonData.id, JSON.stringify(hackathonData))
+  }, [hackathonData])
```
verzano, line 72: so... when there is no value in local storage this will set it, but if there is and hackathonData changes it won't do anything?  Shouldn't we overwrite the value when it changes?

hms-app/src/components/RelevantIdeasLoader.tsx
```diff
@@ -39,52 +39,44 @@ export default function RelevantIdeasLoader({
...
 
   useEffect(() => {
     setRelevantIdeaList((relevantIdeaList) => {
       return []
     })
     loadRelevantIdeaDetails()
+    setHackathon(hackathonData)
   }, [hackathonData])
```
verzano, line 80: @KBernaszuk does this make a loop?

hms-app/src/components/RelevantIdeasLoader.tsx
```diff
@@ -103,9 +95,11 @@ export default function RelevantIdeasLoader({
 
   useEffect(() => {
     setRelevantIdea(relevantIdeaList)
```
verzano, line 97: this looks like it passes a list of ideas to something that expects one idea?

verzano (approved):

PR #396, opened 2022-06-03
refactored idea card buttons

### LINKED ISSUE #395: refactor IdeaDetails
In IdeaDetails in the Accordion the only real changes are the buttons so refactor it

LetsPoke:
closes #395

### DIFF
--- a/hms-app/src/components/card-details/IdeaDetails.tsx
+++ b/hms-app/src/components/card-details/IdeaDetails.tsx
@@ -236,37 +236,22 @@ export default function IdeaDetails(props: IProps) {
                     </Group>
                   )}
 
-                  {type === IdeaCardType.Owner && (
-                    <Group position="left" mt="xl">
-                      {deleteModal}
-                      <Button
-                        color={'red'}
-                        onClick={() => setDeleteModalOpened(true)}
-                      >
-                        Delete
-                      </Button>
-                      {editModal}
-                      <Button onClick={() => setEditModalOpened(true)}>
-                        Edit
-                      </Button>
-                    </Group>
-                  )}
-
-                  {type === IdeaCardType.Admin && (
-                    <Group position="left" mt="xl">
-                      {deleteModal}
-                      <Button
-                        color={'red'}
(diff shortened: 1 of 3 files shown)


hms-app/src/components/card-details/IdeaDetails.tsx
```diff
@@ -217,42 +217,26 @@ export default function IdeaDetails(props: IProps) {
...
-                  >
-                    <div>{ideaDetails()}</div>
+                  <div>{ideaDetails()}</div>
 
+                  {type === IdeaCardType.IdeaPortal && (
                     <Group mt="xs" position={'right'} style={{ paddingTop: 5 }}>
                       <Button variant={'outline'} color={'yellow'}>
                         Add to Favorites
                       </Button>
                       <Button>Participate</Button>
```
verzano, line 235: should this only say participate if it is not you idea?

verzano: in fact couldn't the design depend purely only who you are in relation to the Idea?  If you are an admin or the owner of the Idea then it should show the Edit/Delete buttons and if you are not then it should show Participate?

And the participate button should show something different depending on if you are already participating in the Idea right?

verzano: Should the Participate button always be there even?

verzano: And lastly I'm always more of a fan of using icons/glyphs instead of text for buttons.  It allows them to be consistently sized for all languages and means you don't need to bother translating

LetsPoke: I see your point and it somehow makes sense 

not 100% sure if the "role" dependencie would be that much better

verzano: well it would mean that only the area of the buttons would need to be dynamic.  It will also ensure that whenever/wherever I look at an Idea description it will always appear the same to me

verzano (approved):

PR #424, opened 2022-06-27
357 store last selected hackathon

jakepanter:
- added Redux
- UI changes
- visual error message
- other minor changes


hms-api/src/repository/dynamo-db.ts
```diff
@@ -7,7 +7,8 @@ export function getClient(): DynamoDBClient {
     const endpoint = 'http://localhost:8000';
     return new DynamoDBClient({endpoint: endpoint});
   } else {
-    return new DynamoDBClient({});
+    const endpoint = 'http://localhost:8000';
+    return new DynamoDBClient({endpoint: endpoint});
```
verzano, line 11: doesn't this kinda defeat the purpose :thinking:

hms-app/src/actions/HackathonActions.tsx
```diff
@@ -1,8 +1,13 @@
...
+  Hackathon,
+  HackathonPreview,
+  parseHackathon,
+  parseHackathons,
+} from '../common/types'
 
 const core_url = process.env.REACT_APP_CORE_URL
 
-export const getListOfHackathons = (): Promise<Hackathon[]> => {
+export const getListOfHackathons = (): Promise<HackathonPreview[]> => {
```
verzano, line 10: but `parseHackathons` returns a `Hackathon[]`?  I would think we need another function

LetsPoke: I think parseHackathons then should probably return HackathonPreview? Every list only return previews right?

hms-app/src/common/types.ts
```diff
@@ -47,6 +47,17 @@ export type Hackathon = {
...
 
+export type HackathonSerializable = {
+  id: string
+  title: string
+  startDate: string
+  endDate: string
+  participants?: ParticipantPreview[]
+  categories?: CategoryPreview[]
+  ideas?: IdeaPreview[]
+  status?: HackathonStatus[]
```
LetsPoke, line 58: where does this status come from?

hms-app/src/pages/HomeQandAContent.tsx
```diff
@@ -0,0 +1,42 @@
+export const qAndAList = [
```
LetsPoke, line 1: not a page right? so maybe should be moved to the common folder or something like this

hms-app/src/actions/HackathonActions.tsx
verzano: yes it should have the return type changed, but then the name too.  That way it is obvious it returns `HackathonPreview[]` and not `Hackathon[]`

verzano (approved):

Issue #439, opened 2022-06-29
placeholder hackathon on pages


LetsPoke:
remove placeholder hackathons when nothing is selected in idea portal and your ideas page

should load the last selected hackathon

probably will be solved with issue https://github.com/knoffhoff/hms/issues/437

PR #447, opened 2022-07-04
removed localstorage

### LINKED ISSUE #437: replace localstorage with redux
replace localstorage with redux on all pages and make sure loading works when something get changed

LetsPoke:
for now removed localstorage 

closes #437 #439


hms-app/src/components/card-details/HackathonDetails.tsx
```diff
@@ -261,7 +261,7 @@ export default function HackathonDetails(props: IProps) {
         </div>
       )}
 
-      {hackathonData.startDate &&
```
verzano, line 264: this looks really hacky, and it isn't clear from the code how `> new Date(1)` is a valid check :thinking:

LetsPoke: would be > new Date(1970-01-01T00:00:00.000Z) better? somehow i couldnt check for if title != '' or somethink like that so i thought i check for date not for one of the strings

verzano: thats not really good either, we should work it so that if startDate is null or undefined then it works and if it ever has a value then it is valid :thinking:

LetsPoke: sounds like a good compromise

hms-app/src/pages/IdeaPortal.tsx
```diff
@@ -144,22 +136,20 @@ function IdeaPortal() {
...
-              : addHackathonParticipant
-          }
-          color={participantCheck ? 'red' : 'blue'}
-        >
-          {participantCheck ? 'Leave Hackathon' : 'Join Hackathon'}
-        </Button>
-      )}
-
-      {!isLoading && (
+      {!isLoading && hackathonData.startDate > new Date(1) && (
```
verzano, line 139: here you should use the constant

hms-app/src/pages/YourIdeas.tsx
```diff
@@ -56,32 +56,32 @@ export default function YourIdeas() {
...
       <RelevantIdeasLoader
-        setHackathon={setHackathon}
+        setHackathon={setHackathonData}
         setRelevantIdeas={setRelevantIdeas}
         selectedHackathonId={selectedHackathonId}
         setLoading={setIsLoading}
       />
 
-      {!isLoading && (
+      {!isLoading && hackathonData.startDate > new Date(1) && (
```
verzano, line 65: here too

hms-app/src/pages/YourIdeas.tsx
```diff
@@ -56,32 +56,32 @@ export default function YourIdeas() {
...
         <div>
-          <h2>{hackathon.title}</h2>
+          <h2>{hackathonData.title}</h2>
           <h2>
-            Start Date: {new Date(hackathon.startDate).toDateString()} End Date:{' '}
-            {new Date(hackathon.endDate).toDateString()}
+            Start Date: {new Date(hackathonData.startDate).toDateString()} End
+            Date: {new Date(hackathonData.endDate).toDateString()}
```
verzano, line 70: I would think this should be done the same way as on the `home` page.  In fact you should extract this to a function (either one that takes in a Hackathon or is part of the Hackathon's defintion) and spits out a nice consistent string

hms-app/src/components/card-details/HackathonDetails.tsx
```diff
@@ -261,7 +262,7 @@ export default function HackathonDetails(props: IProps) {
...
+      {hackathonData.startDate != NULL_DATE &&
```
verzano, line 265:
```suggestion
      {hackathonData.startDate !== NULL_DATE &&
```

verzano: and the same thing in the other places too

LetsPoke: updated

verzano (approved):

PR #448, opened 2022-07-04
390 limit the idea card title

### LINKED ISSUE #390: limit the idea card title
set a input limit for the idea title when creating an idea, so that its always gets displayed in a nice way on the idea cards

LetsPoke:
closes #390

### DIFF
--- a/hms-app/src/components/input-forms/IdeaForm.tsx
+++ b/hms-app/src/components/input-forms/IdeaForm.tsx
@@ -52,6 +52,7 @@ function IdeaForm(props: IProps) {
     goal: '',
     creationDate: new Date(),
   })
+  const allowedIdeaTitleLength = 100
 
   const setIdea = () => {
     if (idea) {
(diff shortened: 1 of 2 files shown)


hms-app/src/components/input-forms/IdeaForm.tsx
```diff
@@ -52,6 +52,7 @@ function IdeaForm(props: IProps) {
     goal: '',
     creationDate: new Date(),
   })
+  const allowedIdeaTitleLength = 100
```
verzano, line 55: it would be best to define this in a single place so that it is hard to make the mistake of having two values.  Maybe somewhere close to the Idea interface since it is tied to that.

verzano: also maybe 128 :P nice power of 2

hms-app/src/components/input-forms/IdeaForm.tsx
```diff
@@ -202,6 +206,11 @@ function IdeaForm(props: IProps) {
               <Textarea
                 label="Title"
                 required
+                error={
+                  ideaText.title.length > allowedIdeaTitleLength
+                    ? 'max 100 words'
```
verzano, line 211: the max is 100 chars not words :P and instead of "100" use a reference to the max length value

LetsPoke: updated

verzano (approved):

PR #452, opened 2022-07-05
436 add highlight color to the header menu

### LINKED ISSUE #436: add highlight color to the header menu
add highlight color to the header menu to show which site is active

LetsPoke:
closes #436


hms-app/src/common/constants.ts
```diff
@@ -1 +1,3 @@
 export const NULL_DATE = new Date('1970-01-01')
+export const HeaderActiveColor = 'lightblue'
```
verzano, line 2: so I think there is already a `styles` file that this wold fit more logically in

verzano: also the capitalization is probably wrong, constants should be ALL_CAPS or possibly camalCase

hms-app/src/components/HeaderMenu.tsx
```diff
@@ -18,10 +19,21 @@ export default function HeaderMenu({ links }: HeaderSearchProps) {
...
+        key={link.label}
+        to={link.link}
+        className={classes.link}
+        style={{
+          backgroundColor:
+            location.pathname.slice(1) === link.link
+              ? HeaderActiveColor
+              : undefined,
+        }}
+      >
```
verzano, line 36: all of this should probably be defined in the `styles` file

verzano: could make a new class called... headerLink or something...

LetsPoke: yeah but then you would to pass in location.pathname and link.link?

verzano (approved):

PR #498, opened 2022-07-25
429 add idealo styling

### LINKED ISSUE #429: add idealo styling
- create a styling configuration for idealo CI that can be used as a template for others
- it should be reusable and not hardcoded styles
- add logo on navbar

LetsPoke:
added idealo colors 
added colors file 
changed header color

closes #429


hms-app/src/components/card-details/CategoryDetails.tsx
```diff
@@ -70,7 +71,10 @@ export default function CategoryDetails(props: IProps) {
       </Text>
       <Text className={classes.title}>Title: {categoryData.title}</Text>
       {!isLoading && (
-        <Button color={'red'} onClick={() => deleteSelectedCategory()}>
+        <Button
+          style={{ backgroundColor: LEAVE_BUTTON_COLOR }}
```
verzano, line 75: should probably be `DELETE_BUTTON_COLOR` (I know it doesn't exist yet but maybe should)

verzano: or I guess this is more... a confirm button :thinking:

verzano: basically my point is that if you have colors with names specific to where they are used then that color should only be used there.  And then when you have other places you should have other named colors

LetsPoke: what you think about RED_BUTTON as name?

LetsPoke: or split between leave and delete button?

jakepanter (approved):

PR #499, opened 2022-07-25
Move backend infrastructure to cdk

jakepanter:
moved backend resources from serverless to CDK


LetsPoke (approved):

hms-api/serverless.yml
```diff
@@ -141,7 +141,7 @@ functions:
     handler: src/handler/idea/list.list
     events:
       - http:
-          path: hackathon/{hackathonId}/ideas
+          path: hackathon/{id}/ideas
```
verzano, line 144: this will certainly break tests, and I would actually say to go in the opposite direction and use the longer names for everything if possible.  It's fine like this though, but fix the tests :)

jakepanter: Cloudformation would not accept the longer name though, if there is another path with `hackathon/{id}*`. That was the issue with deployment.

verzano: weird

hms-api/__tests__/handler/category/list.test.ts
```diff
@@ -1,24 +1,25 @@
-import {randomCategory} from '../../repository/domain/category-maker';
-import {list} from '../../../src/handler/category/list';
-import Uuid, {uuid} from '../../../src/util/Uuid';
-import NotFoundError from '../../../src/error/NotFoundError';
-import CategoryListResponse from '../../../src/rest/CategoryListResponse';
-import * as categoryService from '../../../src/service/category-service';
+import { randomCategory } from "../../repository/domain/category-maker";
```
verzano, line 1: :cry:

jakepanter: sorry :D the formatting was done on the other PR

PR #549, opened 2022-08-02
548 FE[yli] added category to idea detail

yankang-is24:
Screenshot 2022-08-02 at 16 25 04


LetsPoke:
probably the category should look same like the skills? 
so you have a badge where you see the category without opening the details?

then you can see on the first look what skills are needed and which category the idea belongs to? 
open for discussion

yankang-is24:
i thought about this too, the question is, is `category` so important? do people choose an idea based on the category? i don't think so. So i put it under `details`. I'm open for other ideas :)

LetsPoke:
kinda unsure 
I guess its good on the details for now and if people starts to complain we can change it

LetsPoke (approved):

jakepanter (approved):

PR #566, opened 2022-08-08
Added category and participants to pitch slide

jakepanter:


hms-app/src/components/PitchTimer.tsx
```diff
@@ -42,19 +50,57 @@ const useStyles = createStyles((_theme, _params, getRef) => ({
...
+  progress: {
+    position: 'fixed',
+    top: 0,
+    left: 0,
+    width: '100%',
+  },
 }))
 
 const getMillis = (minutes: number, seconds: number) => {
   return minutes * 60 * 1000 + seconds * 1000
```
verzano, line 89: it is a stupidly small improvement (and actually whatever transpiles the TS might do it already), but you should just multiple by 60000 instead of doing the additional multiplication here.  There is probably some bit shifting operation to do the multiplication that would be even faster but that would just be excessive

hms-app/src/components/PitchTimer.tsx
```diff
@@ -42,19 +50,57 @@ const useStyles = createStyles((_theme, _params, getRef) => ({
...
 }
 
+const getSeconds = (millis: number) => {
+  return Math.floor(millis / 1000)
+}
+
 function PitchTimer(time: PitchTimerProps) {
```
verzano, line 96: `time` isn't really a good name for this variable.  Just `props` would be better (since it does contain more than time information).

hms-app/src/components/PitchTimer.tsx
```diff
@@ -42,19 +50,57 @@ const useStyles = createStyles((_theme, _params, getRef) => ({
...
   const timeInMillis = getMillis(time.minutes, time.seconds)
   const { classes } = useStyles()
   const [date, setDate] = useState(Date.now() + timeInMillis)
   const [countdownApi, setCountdownApi] = useState<CountdownApi>()
   const [, forceUpdate] = useReducer((x) => x + 1, 0)
   const [progress, setProgress] = useState(0)
+  const [timeLeft, setTimeLeft] = useState(timeInMillis)
```
verzano, line 103: `millisLeft` could be a better name for this (gives more context)

hms-app/src/components/PitchTimer.tsx
```diff
@@ -110,6 +157,12 @@ function PitchTimer(time: PitchTimerProps) {
 
   return (
     <div className={classes.outerBox}>
+      <Progress
+        className={classes.progress}
+        value={progress}
+        color={isCompleted() ? 'red' : 'gray'}
```
verzano, line 163: using colors defined outside this file might be better (allow for easier customization and ensures we use the same colors for similar things)

hms-app/src/components/PitchTimer.tsx
```diff
@@ -150,11 +203,19 @@ function PitchTimer(time: PitchTimerProps) {
...
-      />
+      {timeLeft > 0 && timeLeft < 4000 && (
+        <div
+          className={`${classes.fullscreenOverlay} ${
+            timeLeft > 1000 ? classes.bgOrange : classes.bgRed
+          }`}
+        >
+          <Center>
+            <span className={classes.fullscreenCounter}>
+              {getSeconds(timeLeft)}
```
verzano, line 214: these seems a bit odd, we pass in the remaining seconds to the function, then convert them (and the minutes) to a millisecond value and then here we convert that back to seconds :thinking:

But this could be > 60 right? so you could have like 450 seconds

in fact do we EVER need millisecond precision here? can't we just store remaining seconds?

verzano: ah no, it will count down from 4 seconds, so never more than 60.

hms-app/src/components/PitchTimer.tsx
jakepanter: these are MantineColor identifiers so are basically keys for MantineColors that are defined in the theme

hms-app/src/components/PitchTimer.tsx
jakepanter: true, but I could refactor it a bit. Just need the millis to be added to the Date.now()

LetsPoke (approved):

PR #795, opened 2022-10-24
518 your ideas list doesnt refresh automatically after creating a new idea

### LINKED ISSUE #518: "your ideas" list doesnt refresh automatically after creating a new idea

### LINKED ISSUE #555: Idea should be refreshed after a user joins an idea

LetsPoke:
update participant list after join/leave idea
update idealist after creating idea

closes #555 
closes #518


hms-app/src/components/input-forms/IdeaForm.tsx
```diff
@@ -239,6 +244,12 @@ function IdeaForm(props: IProps) {
     loadAvailableCategories()
   }, [hackathon])
 
+  useEffect(() => {
+    if (props.loaderState) {
+      props.loaderState(loader)
+    }
+  }, [loader])
```
verzano, line 251: not sure about this, feels like the if condition is wrong 🤔

LetsPoke: should be right
checks if the prop loaderstate exist 
and if so return this function with the actual value of the loader

verzano: ok so this was just a bit of my misunderstanding of the code :P it works the way it should

verzano (approved):

PR #1068, opened 2022-12-13
982 ideas should be independent from hackathons

### LINKED ISSUE #646: allow Hackathons with NULL date
add the possibility to create Hackathon without giving them start and end date

LetsPoke:
- adds a new all ideas endpoint
- updated the header with the split between hackathons and ideation portal
- created a raw version of the ideation portal

closes #646


hms-app/src/components/HeaderMenu.tsx
```diff
@@ -30,14 +31,25 @@ interface HeaderSearchProps {
   links: {
     link: string
     label: string
-    links?: { link: string; label: string }[]
+  }[]
+  hackLinks: {
```
jakepanter, line 35: naming could be more specific: sth like hackathonLinks

jakepanter (approved):

hms-api/src/repository/domain/Hackathon.ts
```diff
@@ -79,15 +83,11 @@ class Hackathon {
...
-    if (!this.startDate) {
-      result.addFailure('startDate is null');
-    }
-
-    if (!this.endDate) {
-      result.addFailure('endDate is null');
-    }
-
-    if (this.startDate >= this.endDate) {
+    if (
```
verzano, line 86: this is shorter, yes, but you lose a bit of information this way

verzano: I'm guessing you did this because the `if (this.startDate >= this.endDate)` might not work right when either `this.startDate` or `this.endDate` is null.  In that case you might want to change just that one if statement

verzano: ah now I see, because startDate and endDate can be null... hmmm... I think that isn't a good change, because now it is possible to make a Hackathon with no start or end date

hms-api/src/repository/hackathon-repository.ts
```diff
@@ -97,8 +97,8 @@ function itemToHackathon(item: {[key: string]: AttributeValue}): Hackathon {
   return new Hackathon(
     item.title.S,
     item.description.S,
-    new Date(item.startDate.S),
-    new Date(item.endDate.S),
+    item.startDate ? new Date(item.startDate.S) : null,
+    item.endDate ? new Date(item.endDate.S) : null,
```
verzano, line 101: this too, seems like maybe not a good change

hms-api/src/repository/idea-repository.ts
```diff
@@ -44,6 +44,21 @@ export async function listIdeasForHackathon(
   );
 }
 
+export async function listIdeasAll(): Promise<Idea[]> {
```
verzano, line 47: this is a VERY risky thing to do... eventually this will become VERY large, in fact, dynamo won't return all of them.  It will return a paged response that you would need to iterate over.  That is actually true for basically all of the requests we are making, but when you limit it based on some query parameter it becomes less of a concern

hms-api/__tests__/repository/domain/Hackathon.test.ts
```diff
@@ -10,30 +10,24 @@ describe('Hackathon Validation', () => {
     ['Empty ID', makeHackathon({id: ''} as HackathonData), true],
     ['Null Title', makeHackathon({title: null} as HackathonData), true],
     ['Empty Title', makeHackathon({title: ''} as HackathonData), true],
-    ['Null Start Date',
-      makeHackathon({startDate: null} as HackathonData),
-      true],
-    ['Null End Date',
-      makeHackathon({endDate: null} as HackathonData),
-      true],
```
verzano, line 18: these are for sure still possible states and need to be tested

hms-api/src/rest/HackathonPreviewResponse.ts
```diff
@@ -49,19 +49,21 @@ class HackathonPreviewResponse {
     a: HackathonPreviewResponse,
     b: HackathonPreviewResponse,
   ): number {
-    let diff = a.startDate.getTime() - b.startDate.getTime();
+    if (a.startDate && b.startDate && a.endDate && b.endDate) {
```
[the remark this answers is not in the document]
LetsPoke: unsure how to solve this 

maybe we can put an else behind this if so if it doesnt go into then it atleast could return a hardcoded value in the else?

PR #1137, opened 2023-01-03
1102 update ideacard for hackathon change

### LINKED ISSUE #517: add default category to the backend
as you need a hackathon category to submit ideas we should create atleast 1 default category automatically when creating a new hackathon (in case the admins forgot to create one afterwards) somethink like "main categoy"

### LINKED ISSUE #1096: Find Solution for Categories
If we want to move ideas from one hackathon to another we need to find a solution for the categories
what should happen with the idea category when changing hackathon? 
do all hackathons have the same ideas?
will categories get resettet when change hackathon? 

part of #982

### LINKED ISSUE #1102: update ideaCard for hackathon change
1. IdeaForm for new idea in IdeaPortal should not contain fields for category and skills (check)

2. In the IdeaPortal the ideaCards should not have skills, categories and participants... (check)

3. User should be able to move an idea to a Hackathon in the IdeaForm (check)

4. When idea gets moved automaticaly remove idea from old hackathon and add idea to new hackathon (check)

5. When moved from ideaPortal to Hackathon user should be asked to add skills and category (check)

6. When move from Hackathon to IdeaPortal skills, category and participants should be resetted (check)

LetsPoke:
- Updated IdeaForm for idea creation
- Updated IdeaCards
- Added MoveIdea button
- Ask user to select IdeaCategory when switching Hackathon
- When move Idea remove IdeaParticipants

closes #1102 
closes #1096 
closes #517


hms-api/src/handler/hackathon/create.ts
```diff
@@ -15,6 +16,10 @@ export async function create(event, context, callback) {
       request.endDate,
     );
 
+    if (hackathon) {
+      await createCategory('General', 'General', hackathon.id);
+    }
```
verzano, line 21: createHackathon should never return nothing, checking here is a bit redundant.  Also it would be good to leave a comment here as to why we create a `General` category

verzano: I would also say that this should be done in the service, and not in the controller.  The controller should pretty dumb and not need to know that the hackathon requires this default category.  I would also argue that the Hackathon is invalid until this category exists so it would be best if it doesn't return from the service until it does

verzano: so... I would say move this call into the service.  And then for the tests for the service you will need to mock this call by "spy"ing on it and then also ensure it is called with the correct values.

hms-api/src/service/hackathon-service.ts
```diff
@@ -35,8 +36,17 @@ export async function createHackathon(
     throw new ValidationError(`Cannot create Hackathon`, result);
   }
 
+  const category = new Category('General', 'General', hackathon.id);
+  const categoryResult = category.validate();
+  if (categoryResult.hasFailed()) {
+    throw new ValidationError(`Cannot create general Category`, categoryResult);
+  }
```
verzano, line 43: you should also store this category in the DB I would imagine 🤔

verzano: ah I should have kept reading, I see you do :D

hms-api/src/handler/hackathon/create.ts
```diff
@@ -16,10 +16,6 @@ export async function create(event, context, callback) {
       request.endDate,
     );
 
-    if (hackathon) {
```
verzano: line 6 is no longer used

hms-api/src/service/hackathon-service.ts
verzano: hmmm now that I think about it a bit more, it might be best to encapsulate all of this within the category-service

verzano (approved):

PR #1224, opened 2023-01-30
Feat: 1152 add commentary function to ideas

LetsPoke:
added commentary function 
backendwise comments have a "ReplyTo" field to connect them
but for now frontendwise users are only able to write high level comments with no "ReplyTo" id


hms-api/serverless.local.yml
```diff
@@ -55,6 +57,8 @@ provider:
     PARTICIPANT_TABLE: participant
     PARTICIPANT_BY_HACKATHON_ID_INDEX: participant-by-hackathonId
     PARTICIPANT_BY_HACKATHON_ID_USER_ID_INDEX: participant-by-hackathonId-userId
+    COMMENT_TABLE: ideaComment
+    COMMENT_BY_IDEA_ID_INDEX: ideaComment-by-ideaId
```
verzano, line 61: Just for consistency I'd do...

```suggestion
    IDEA_COMMENT_TABLE: ideaComment
    IDEA_COMMENT_BY_IDEA_ID_INDEX: ideaComment-by-ideaId
```

verzano: be sure to do a global search for the names and replace them

hms-api/src/handler/comment/create.ts
```diff
@@ -0,0 +1,19 @@
...
+import {createComment} from '../../service/idea_comment-service';
+import {wrapHandler} from '../handler-wrapper';
+import CommentCreateRequest from '../../rest/Comment/CommentCreateRequest';
+import CommentCreateResponse from '../../rest/Comment/CommentCreateResponse';
+
+export async function create(event, context, callback) {
+  await wrapHandler(async () => {
+    const request = CommentCreateRequest.parse(event.body);
+    const comment = await createComment(
+      request.ideaId,
```
verzano, line 11: in the URL you have the ID of the idea as part of the path, but here you also have it as part of the request

hms-api/src/handler/comment/create.ts
```diff
@@ -0,0 +1,19 @@
...
+      request.userId,
```
verzano, line 12: the userID of the person that is creating it should come from the authentication token and not the request (this prevents you from creating comments as someone else)

hms-api/src/handler/comment/create.ts
```diff
@@ -0,0 +1,19 @@
...
+      request.text,
+      request.replyTo,
```
verzano, line 14: replyTo sounds too much (exactly) like the email concept of `REPLY-TO` maybe something more like `parentCommentId` would be good (this would also allow a sort of infinite nesting of comments)

hms-api/src/handler/hackathon/create.ts
```diff
@@ -1,8 +1,8 @@
 import {buildResponse} from '../../rest/responses';
 import {createHackathon} from '../../service/hackathon-service';
 import {wrapHandler} from '../handler-wrapper';
-import HackathonCreateRequest from '../../rest/HackathonCreateRequest';
-import HackathonCreateResponse from '../../rest/HackathonCreateResponse';
+import HackathonCreateRequest from '../../rest/Hackathon/HackathonCreateRequest';
```
verzano, line 4: use lowercase for folders please

hms-api/src/repository/domain/IdeaComment.ts
```diff
@@ -0,0 +1,94 @@
...
+   * the String that contains the text of the Comment
+   */
+  text: string;
+
+  /**
+   * the comment object that this comment is a reply to
+   *
+   * if this is a top level comment, this will be null
+   */
+  replyTo: Uuid;
```
verzano, line 39: this should be nullable
```suggestion
  replyTo: Uuid?;
```

hms-api/src/repository/domain/IdeaComment.ts
```diff
@@ -0,0 +1,94 @@
...
+
+    return result;
+  }
+
+  constructor(userId: Uuid, ideaId: Uuid, text: string, replyTo: Uuid);
+  constructor(
+    userId: Uuid,
+    ideaId: Uuid,
+    text: string,
+    replyTo: Uuid,
```
verzano, line 74:
```suggestion
    replyTo: Uuid?,
```

hms-api/src/repository/domain/IdeaComment.ts
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    if (!this.text) {
+      result.addFailure('text is null or empty');
+    }
+
+    return result;
+  }
+
+  constructor(userId: Uuid, ideaId: Uuid, text: string, replyTo: Uuid);
```
verzano, line 69:
```suggestion
  constructor(userId: Uuid, ideaId: Uuid, text: string, replyTo: Uuid?);
```

hms-api/src/repository/domain/IdeaComment.ts
```diff
@@ -0,0 +1,94 @@
...
+    replyTo: Uuid,
+    id: Uuid,
+    creationDate: Date,
+  );
+
+  constructor(
+    userId: Uuid,
+    ideaId: Uuid,
+    text: string,
+    replyTo: Uuid,
```
verzano, line 83:
```suggestion
    replyTo: Uuid?,
```

hms-api/src/repository/idea_comment-repository.ts
```diff
@@ -0,0 +1,120 @@
+import {
```
verzano, line 1: use a hyphen for `idea-comment`

hms-api/src/repository/idea_comment-repository.ts
```diff
@@ -0,0 +1,120 @@
...
+export async function putComment(comment: Comment): Promise<void> {
+  await dynamoDbClient.send(
+    new PutItemCommand({
+      TableName: process.env.COMMENT_TABLE,
+      Item: {
+        id: {S: comment.id},
+        userId: {S: comment.userId},
+        ideaId: {S: comment.ideaId},
+        text: {S: comment.text},
+        replyTo: {S: comment.replyTo},
```
verzano, line 48: this will have to be changed since replyTo (or whatever the name will be) is nullable

hms-api/src/service/idea_comment-service.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  const comment = new Comment(userId, ideaId, text, replyTo);
+  const result = comment.validate();
+  if (result.hasFailed()) {
+    throw new ValidationError('Cannot create Comment', result);
+  }
+
+  await putComment(comment);
+
+  return comment;
```
verzano, line 45: so a few things about this function:
1. You don't need them as `else if` statements because you throw inside each of them so you don't need to worry about them executing when they shouldn't (I get it makes it a little bit "shorter" like this but I would argue it is more readable as just a bunch of `if` statements)
2. `commentAlreadyExists(new Comment...` will never evaluate to true because when you make a new Comment it gets a new random ID (which for sure won't be in DynamoDB) so I would say this check isn't necessary here.  It WOULD be necessary if we had some other way of determining that two comments are the "same" (like identical text or something).  But as that isn't the case we could drop the check here.
3. You create a `new Comment` multiple times in this function, and they will each have different IDs (making them "different") as well it wastes a bit of memory and cpu cycles in doing so.  It would be better to just do it once, right before the first time you use it.
4. `Comment` as a class doesn't seem to exist anywhere, `IdeaComment` does however (I think you meant to use that)

hms-api/src/service/idea_comment-service.ts
```diff
@@ -0,0 +1,130 @@
...
+  try {
+    user = await getUser(comment.userId);
+  } catch (e) {
+    throw new ReferenceNotFoundError(
+      `Cannot get Comment with id ${id}, ` +
+        `unable to get User with id ${comment.userId}`,
+    );
+  }
+
+  let idea;
```
verzano, line 61: you get this idea, but never use it 🤔

hms-api/src/service/idea_comment-service.ts
```diff
@@ -0,0 +1,130 @@
...
+  //Todo: evaluate how we handle comments with deleted users
+  let users;
+  try {
+    users = await Promise.all(
+      comments.map((comment) => getUser(comment.userId)),
+    );
+  } catch (e) {
+    throw new ReferenceNotFoundError(
+      `Cannot get Comment list, ` +
+        `unable to get User with id ${users.userId}`,
```
verzano, line 92: `users` would be a list with many User objects in it.  And if this were to catch it will still be undefined.  You probably should print the list of ALL users in the comments (`comments.

hms-api/src/service/participant-service.ts
```diff
@@ -110,7 +110,6 @@ export async function removeParticipant(
   id: Uuid,
 ): Promise<ParticipantDeleteResponse> {
   try {
-    await removeIdeasForOwner(id);
```
verzano, line 113: why did this go away? It also seems really unrelated to the comments stuff

hms-api/src/service/idea_comment-service.ts
```diff
@@ -0,0 +1,130 @@
...
+} from '../repository/idea_comment-repository';
+import ValidationError from '../error/ValidationError';
+import Comment from '../repository/domain/IdeaComment';
+import ReferenceNotFoundError from '../error/ReferenceNotFoundError';
+import CommentResponse from '../rest/Comment/CommentResponse';
+import CommentDeleteResponse from '../rest/Comment/CommentDeleteResponse';
+import CommentListResponse from '../rest/Comment/CommentListResponse';
+import ideaComment from '../repository/domain/IdeaComment';
+
+export async function createComment(
```
verzano, line 20: there seems to be no consensus on whether it is a `Comment` or an `IdeaComment`.  It would be good to just use one everywhere

hms-api/__tests__/service/participant-service.test.ts
```diff
@@ -257,76 +268,71 @@ describe('Delete Participant', () => {
...
     expect(mockRemoveParticipantFromIdeas).toHaveBeenCalledWith(id);
     expect(mockDeleteParticipant).toHaveBeenCalledWith(id);
   });
 
   test('Remove Participant from Ideas Fails', async () => {
     const id = uuid();
 
-    mockRemoveIdeasForOwner.mockImplementation(() => {
-    });
+    // mockRemoveIdeasForOwner.mockImplementation(() => {});
```
verzano, line 285: I do very much wonder why all of these have been commented out

hms-api/__tests__/service/idea_comment-service.test.ts
```diff
@@ -0,0 +1,61 @@
...
+jest.spyOn(userRepository, 'getUser').mockImplementation(mockGetUser);
+const mockGetIdea = jest.fn();
+jest.spyOn(ideaRepository, 'getIdea').mockImplementation(mockGetIdea);
+const mockGetComment = jest.fn();
+jest
+  .spyOn(ideaCommentRepository, 'getComment')
+  .mockImplementation(mockGetComment);
+
+describe('Create Idea Comment', () => {
+  test('Missing User', async () => {
```
verzano, line 36: should also write a happy path test as well as the other failure states

hms-api/__tests__/repository/domain/IdeaComment.ts
```diff
@@ -0,0 +1,17 @@
+import {
```
verzano, line 1: this class should be named IdeaComment.test.ts

hms-api/__tests__/repository/domain/IdeaComment.ts
```diff
@@ -0,0 +1,17 @@
...
+  randomIdeaComment,
+} from './ideaComment-maker';
+
+describe('IdeaComment Validation', () => {
+  it.each([
+    ['Happy Path', randomIdeaComment(), false],
+    ['Null ID', makeIdeaComment({id: null} as IdeaCommentData), true],
+    ['Empty ID', makeIdeaComment({id: ''} as IdeaCommentData), true],
+    ['Null User ID', makeIdeaComment({userId: null} as IdeaCommentData), true],
+    ['Empty User ID', makeIdeaComment({userId: ''} as IdeaCommentData), true],
```
verzano, line 13: maybe add a check to ensure that replyTo can be null 🤔

hms-api/__tests__/handler/comment/remove.test.ts
```diff
@@ -0,0 +1,57 @@
...
+
+const mockRemoveComment = jest.fn();
+jest
+  .spyOn(commentService, 'removeComment')
+  .mockImplementation(mockRemoveComment);
+
+describe('Delete Comment', () => {
+  test('Happy Path', async () => {
+    const id = uuid();
+    mockRemoveComment.mockResolvedValue(new UserDeleteResponse(id));
```
verzano, line 15: I think you have the wrong mocked value here

hms-api/src/repository/idea-repository.ts
```diff
@@ -156,6 +156,15 @@ export async function getIdea(id: Uuid): Promise<Idea> {
...
 }
 
+export async function ideaExists(id: Uuid): Promise<boolean> {
+  try {
+    await getIdea(id);
+    return true;
+  } catch (error) {
+    return false;
+  }
+}
```
verzano, line 166: this is a bit of an antipattern.  try/catch should be used for truly exceptional states.  In the case where the idea does not exist (as far as this bit of code is concerned) is not exceptional.  In that case you could easily modify this to not work using a thrown exception and instead just be something like:
```suggestion
export async function ideaExists(id: Uuid): Promise<boolean> {
const output = await dynamoDBClient.send(
    new GetItemCommand({
      TableName: process.env.IDEA_TABLE,
      Key: {id: {S: id}},
    }),
  );
  return !!output.Item;
}
```

Then you leave it up to the caller of this function to determine if the state is "exceptional" or not

hms-app/src/common/styles.ts
```diff
@@ -46,6 +46,12 @@ export const styles = createStyles((theme) => ({
     lineHeight: '1.5em',
   },
 
+  smallText: {
```
verzano, line 49: I feel like there is some way to extend another style... I would think this should just extend `text` and overwrite the properties which should be different

verzano:
One thing I think is also missing from this (I might have missed it) is the deletion of comments when an Idea is deleted as well as when the user that created them is deleted.  In addition if a comment is deleted all comments which pointed to it with their `replyTo` should also be removed (I suppose you could also replace them with a blank comment or something 🤷)

hms-api/src/handler/comment/create.ts
LetsPoke: sorry I understand what you mean here

hms-api/src/handler/comment/create.ts
LetsPoke: I bet this would then be also the case for the other create handler?

hms-api/src/repository/domain/IdeaComment.ts
LetsPoke: Bildschirmfoto 2023-01-31 um 14 08 44

LetsPoke: works with replyTo?: Uuud; but this causes a lot of issues with the order because a optional parameter needs be placed after all needed parameters so have to change a lot und will do this tomorrow

hms-api/src/service/participant-service.ts
LetsPoke: unsure if we really should delete all ideas if a user leaves a hackathon...more like when the user gets deleted or something like that and even then I would argue that we should just set the user to "anonymus" and keep the idea but thats more like a personal opinion

hms-api/__tests__/service/participant-service.test.ts
LetsPoke: was related to the deletion of "removeIdeasForOwner(id);" in the service file

hms-api/serverless.local.yml
```diff
@@ -313,6 +319,39 @@ functions:
...
+          cors: true
+          authorizer: authorize
+  editIdeaComment:
+    handler: src/handler/comment/edit.edit
+    events:
+      - http:
+          path: idea/comment/{id}
+          method: put
+          cors: true
+          authorizer: authorize
```
verzano, line 353: you wrote a get.ts file for idea comments but no get endpoint

hms-api/serverless.local.yml
```diff
@@ -546,5 +585,31 @@ resources:
...
+          - AttributeName: id
+            KeyType: HASH
+        GlobalSecondaryIndexes:
+          - IndexName: ${self:provider.environment.IDEA_COMMENT_BY_IDEA_ID_INDEX}
+            KeySchema:
+              - AttributeName: ideaId
+                KeyType: HASH
+              - AttributeName: id
+                KeyType: RANGE
             Projection:
```
verzano, line 614: it sounds weird, but end the file with a newline please... it's a weird thing about how strings end in C... and it will shut github up about it

hms-api/src/handler/comment/create.ts
verzano: yes, everywhere

hms-api/src/repository/domain/IdeaComment.ts
verzano: ah wait that was kotlin syntax🤦, should be like this:

```suggestion
  replyTo?: Uuid;
```

or even possibly...
```
  replyTo: Uuid | null;
```

but I think the first is better

hms-api/__tests__/service/idea_comment-service.test.ts
```diff
@@ -0,0 +1,61 @@
+import * as userRepository from '../../src/repository/user-repository';
```
verzano, line 1: this file too can get it's name fixed with the hyphen

hms-api/src/repository/idea_comment-repository.ts
verzano: lol, you'd need to check the other spots where we have nullable values but in the end it needs to be something like this when the value is null

```
  replyTo: {NULL: true},
```

I think there is a utility function I wrote somewhere that does this, I know one exists but it might not be in it's own file.  If it isn't in it's own file it would be good to move it for re-use

hms-api/src/service/participant-service.ts
verzano: yeah I'd be ok with that as well, but we need to do something.  If we do nothing then we end up with a lot of broken data (user IDs in the Ideas that point to nothing)

LetsPoke: probably worth a seperate ticket to have a solution for everything when a users gets deleted

hms-api/src/handler/comment/create.ts
verzano: don't we do this already for other things?

hms-api/src/handler/ideaComment/edit.ts
```diff
@@ -0,0 +1,17 @@
...
+export async function edit(event, context, callback) {
+  await wrapHandler(async () => {
+    const id: Uuid = event.pathParameters.id;
+    const request = IdeaCommentEditRequest.parse(event.body);
+
+    await editIdeaComment(id, request.text);
+
+    callback(null, buildResponse(200, new IdeaCommentEditResponse(id)));
+  }, callback);
+}
```
verzano, line 17: I guess we don't really need to change it now, but there is no check in here for permission.  By that I mean we don't check that the user making the edit is the user who created the comment

verzano: the same would also be true for the remove endpoint

hms-api/src/repository/idea_comment-repository.ts
verzano: I'm looking at other bits of the code where I think we should be doing conversions like I mentioned here from one this to null.  I don't see us doing it however.  The only place I really see us doing it is when we have empty lists and we need to store them in DynamoDB.  It might be that this just works no problem 🤔 

... or it will be something we find as a bug later.

Either way I guess it's fine for now

hms-api/serverless.local.yml
verzano: 👁️ 👁️

hms-api/src/rest/ideaComment/IdeaCommentEditRequest.ts
```diff
@@ -0,0 +1,18 @@
+import Uuid from '../../util/Uuid';
+
+class IdeaCommentEditRequest {
+  id: Uuid;
```
verzano, line 4: I don't think you actually use this anywhere since we just get it from the path.  The other editXXXRequest objects don't have them either.

hms-api/src/service/idea-comment-service.ts
```diff
@@ -0,0 +1,129 @@
...
+  //Todo: evaluate how we handle ideaComments with deleted users
+  let users;
+  try {
+    users = await Promise.all(
+      ideaComments.map((ideaComment) => getUser(ideaComment.userId)),
+    );
+  } catch (e) {
+    throw new ReferenceNotFoundError(
+      `Cannot get Comment list, ` +
+        `unable to get User with id ${ideaComments.map((c) => c.userId)}`,
```
verzano, line 89:
```suggestion
        `unable to get Users with IDs: ${ideaComments.map((c) => c.userId)}`,
```

hms-api/__tests__/service/idea_comment-service.test.ts
verzano: 👁️ 👁️

hms-api/src/handler/comment/create.ts
LetsPoke: nope we always use request.userId everywhere

hms-app/src/common/styles.ts
LetsPoke: not sure if this would not overcommplicate it

hms-api/__tests__/service/idea_comment-service.test.ts
LetsPoke: weird authentification token problem when writing happy path and others...maybe you can give me a hint for that in discord

hms-api/src/handler/ideaComment/edit.ts
LetsPoke: https://github.com/knoffhoff/hms/issues/1267

hms-api/src/handler/comment/create.ts
verzano: ok cool, well something we should adjust soon

hms-app/src/common/styles.ts
verzano: the point is more about being able to keep things consistent easily when you change them.  And if the amount of things you set in each of these becomes large then it does make for less code later

hms-api/src/handler/ideaComment/edit.ts
verzano: awesome 👍

hms-api/__tests__/service/idea_comment-service.test.ts
verzano: I doubt it is anything to do with the auth token as we never actually use the rest layer in these unit tests here.  I do see an issue with the `edit.test.ts` file where you call `new ideaCommentEditRequest()`.  You pass in both the id and the text to it, but it should only get the id

LetsPoke: will add missing test tommorrow

hms-api/__tests__/service/idea-comment-service.test.ts
```diff
@@ -0,0 +1,147 @@
...
+
+describe('Create Idea Comment', () => {
+  test('Happy Path', async () => {
+    mockUserExists.mockResolvedValue(true);
+    mockIdeaExists.mockResolvedValue(true);
+
+    const expected = randomIdeaComment();
+
+    expect(
+      await createIdeaComment(expected.ideaId, expected.userId, expected.text),
```
verzano, line 54: so since there are two versions of this function, one with the parentCommentId and one without you should really have two tests for it

PR #1231, opened 2023-01-31
backend changes to add Deeplinks

LetsPoke:


hms-api/serverless.yml
```diff
@@ -43,6 +43,8 @@ provider:
     AZURE_AD_ISSUER: ${file(./secrets.json):AZURE_AD_ISSUER}
     AZURE_AD_CLIENT_PUBLIC_KEY: ${file(./public_key)}
     HACKATHON_TABLE: hackathon
+    HACKATHON_BY_TITLE_INDEX: hackathonById
```
verzano, line 46: I think you meant to do:
```suggestion
    HACKATHON_BY_TITLE_INDEX: hackathonByTitle
```

hms-api/src/repository/hackathon-repository.ts
```diff
@@ -32,12 +34,39 @@ export async function listHackathons(): Promise<Hackathon[]> {
 }
 
 export async function putHackathon(hackathon: Hackathon) {
+  if (await hackathonExists(hackathon.id)) {
+    throw new InvalidStateError('Cannot create Hackathon, it already exists');
+  }
+
+  if (await hackathonSlugExists(hackathon.slug)) {
+    throw new InvalidStateError('Cannot create Hackathon, slug already exists');
+  }
```
verzano, line 43: these are only an issue if you are trying to create a hackathon, but not when you want to edit one so they shouldn't go here

hms-api/serverless.local.yml
```diff
@@ -43,6 +43,8 @@ provider:
     AZURE_AD_ISSUER: ${file(./secrets.json):AZURE_AD_ISSUER}
     AZURE_AD_CLIENT_PUBLIC_KEY: ${file(./public_key)}
     HACKATHON_TABLE: hackathon
+    HACKATHON_BY_TITLE_INDEX: hackathonById
```
verzano, line 46: here too it's wrong

hms-api/serverless.yml
verzano: also, this by it's nature introduces a new requirement.  That hackathon Titles be globally unique, which means you need to validate this on creation and editing to ensure that you can still find a unique hackathon by title

verzano: same thing about the slugs, it now appears that they too must be unique 🤔

verzano: I'm basing this on how I imagine they are meant to be used, if I try to navigate to something like `www.hackathon.com/slug/hack4/` and I gave two things the slug `hack4` it won't know where to go

hms-api/src/repository/hackathon-repository.ts
verzano: now that I've read more, this comment only applies to the `hackathonExists` one.  But likely you need to check that the hackathon title exists... I'll have to dive deeper into this though]

hms-api/src/repository/hackathon-repository.ts
```diff
@@ -32,12 +34,39 @@ export async function listHackathons(): Promise<Hackathon[]> {
 }
 
 export async function putHackathon(hackathon: Hackathon) {
+  if (await hackathonExists(hackathon.id)) {
```
verzano, line 37: I would say the same thing about this as I did about the IdeaComment that when you create a new Hackathon object the Uuid will be generated then and for sure (statistically very close to 0) won't match any existing ones so the check is a bit overkill.  And in the case where you are editing the hackathon needs to exist for you to modify it

hms-api/src/repository/hackathon-repository.ts
```diff
@@ -32,12 +34,39 @@ export async function listHackathons(): Promise<Hackathon[]> {
...
+        endDate: {S: hackathon.endDate.toISOString()},
+        id: {S: hackathon.id},
+        creationDate: {S: hackathon.creationDate.toISOString()},
+        votingOpened: {BOOL: hackathon.votingOpened},
+      },
+    }),
+  );
+}
+
+export async function updateHackathon(hackathon: Hackathon) {
```
verzano, line 62: ok I see what you did now, so the way everything was working was that the repository classes are meant to be really dumb.  They just convert a service layer entity (Hackathon, Idea, User, whatever...) into something DynamoDB understands and makes a request to it.  All of the business logic around checking if something exists already or not is done in the service classes

verzano: my opinion would be to keep it consistent and have the repository classes be dumb and put the smarts in the service layer.  This from a design standpoint has the advantage that if you were say trying to switch from DynamoDB, Redis, or RDS then you don't need to duplicate this business logic in other classes.

This allows for you to have less code overall and be more agile if you wanna change storage options or use multiple for things

hms-api/serverless.yml
verzano: and now that I've looked more you don't seem to actually use the hackathon by title index anywhere

verzano (changes requested):
I think you can drop that one index, and yeah add the tests :P

hms-api/serverless.yml
LetsPoke: yeah I have to remove the title index again... wanted to check title and slug for uniqueness but half way through it jakob convinced me that slug would be enough and you could have 2 hackathons with the same name

hms-api/src/repository/hackathon-repository.ts
LetsPoke: so your saying dynamodb automatically checks that it never randomly choose the same uuid?

verzano: not exactly... so if we consider the flow for creating a new hackathon (or any entity really):

1. request comes into the rest layer with data
2. data is passed to the service layer where a new hackathon is created
  a. the constructor for that hackathon sets new random UUID as the ID of the hackathon
3. the hackathon is checked for validity
4. the hackathon is passed to the repository layer
  a. at this point we can be relatively sure that the ID does not exist in DynamoDB because it is a random UUID
6. we add the hackathon into DynamoDB

so for the case where we make a new one we don't really need to check because of how UUIDs work, we CAN check but it is a bit overkill tbh.  Here is a nice quote about UUIDs:

> [As per Wikipedia](https://en.wikipedia.org/wiki/Universally_unique_identifier), the number of UUIDs generated to have > atleast 1 collision is 2.71 quintillion.
> This is equivalent to generating around 1 billion UUIDs per second for about 85 years.

If we then consider the flow for updating it is mostly the same, and in that flow we want that ID to be in DynamoDB already so that we can update it's values.  This way we can have a single function that we call in the repository to fulfill both purposes.

It's always good to try and get re-use out of a function when possible (and within reason).  If you have to start adding tons of arguments to the function so that it can do all sorts of things, that is a bit of a code smell.  But in these two scenarios it really just has to do the same thing so we can get re-use out of it

verzano (approved):

PR #1252, opened 2023-02-06
Frontend deeplink to hackathons

LetsPoke:
Frontend changes for the DeepLinks


hms-app/src/actions/HackathonActions.tsx
```diff
@@ -43,13 +43,15 @@ export const createHackathon = async (
   instance: IPublicClientApplication,
   title: string,
   DescriptionValue: string,
```
verzano, line 45: this should really be lowercase

hms-app/src/actions/HackathonActions.tsx
```diff
@@ -43,13 +43,15 @@ export const createHackathon = async (
...
   startDate: Date,
   endDate: Date
 ) => {
   const idToken = await getIdToken(instance)
   const options = buildFetchOptions('POST', idToken, {
     title,
     description: DescriptionValue,
+    slug,
     startDate: startDate.toString(),
     endDate: endDate.toString(),
```
verzano, line 56: when you build a type like this it's important to name the parameters for clarity and safety
```suggestion
    title: title,
    description: DescriptionValue,
    slug: slug,
    startDate: startDate.toString(),
    endDate: endDate.toString(),
```

hms-app/src/actions/HackathonActions.tsx
```diff
@@ -65,6 +67,7 @@ export const editHackathon = async (
   hackathonID: string,
   title: string,
   DescriptionValue: string,
```
verzano, line 69: same here about camelCase

hms-app/src/common/constants.ts
```diff
@@ -1,5 +1,6 @@
 import logo from '../assets/idealo-logo.png'
 
 export const NULL_DATE = new Date('1970-01-01')
-export const VALID_DATE = new Date('2000-01-01')
+export const MIN_DATE = new Date('1971-01-01')
+export const MAX_DATE = new Date('2039-12-31')
```
verzano, line 5: why is this the "max" date?

verzano: at the very least comments explaining the rationale behind all of these three constants would be very helpful

hms-app/src/common/redux/hackathonSlice.ts
```diff
@@ -12,6 +12,7 @@ const initialState: HackathonState = {
     {
       id: '33a892ba-8a9e-4edf-947a-f421c86a14b5',
       title: '',
+      slug: '',
```
verzano, line 15: on the backend an empty slug would be invalid (since we check that when we validate a hackathon).  You should give all of these values to ensure they are all valid

verzano: the same thing is true about titles, an empty title is invalid so all of them should get titles as well

hms-app/src/components/HackathonHeader.tsx
```diff
@@ -21,25 +21,29 @@ export default function HackathonHeader(props: IProps) {
...
-              : { from: '#ed6ea0', to: '#ec8c69', deg: 35 }
-          }
-          radius={'xs'}
-        >{`${new Date(
-          hackathonData.startDate
-        ).toLocaleDateString()} - ${new Date(
-          hackathonData.endDate
-        ).toLocaleDateString()}`}</Badge>
-      </Center>
+      {hackathonData.title && (
```
verzano, line 24: in the backend we ensure that a title is never empty so you could remove this check 🤔

verzano: that, of course, means you need to ensure that any other source of the data also is correct

hms-app/src/components/HackathonSelectDropdown.tsx
```diff
@@ -52,7 +63,7 @@ export default function HackathonSelectDropdown({
         return hackathonList
           .filter(
             (hackathon) =>
-              hackathon.endDate < today && hackathon.endDate > VALID_DATE
+              hackathon.endDate < today && hackathon.endDate > MIN_DATE
```
verzano, line 66: why exactly do we do this check? do we expect there to be dates older than that? or do we want to have a configurable MIN_DATE that we can set to "hide" old hackathons?

hms-app/src/components/input-forms/IdeaForm.tsx
```diff
@@ -358,20 +358,20 @@ function IdeaForm(props: IProps) {
...
-                        ? JOIN_BUTTON_COLOR
-                        : dark2,
-                    }}
-                    disabled={buttonIsDisabled}
-                    onClick={createThisIdea}
-                  >
-                    Create
-                  </Button>
-                ))}
+              {(context === 'new' ||
```
verzano, line 361: context should be some kind of enum instead of a raw string value

hms-app/src/pages/IdeationPortal.tsx
```diff
@@ -30,7 +30,7 @@ function IdeationPortal() {
   const loadHackathons = () => {
     getListOfHackathons(instance).then((data) => {
       const upcomingHackathon = data.find(
-        (hackathon) => hackathon.startDate < VALID_DATE
+        (hackathon) => hackathon.startDate < MIN_DATE
```
verzano, line 33: if the start date is BEFORE the MIN_DATE then it would be a very old (and likely invalid) hackathon

hms-app/src/components/HackathonSelectDropdown.tsx
LetsPoke: yeah so the min date hides the ideation portal hackathon and the max dates cutes the date out of the dropdown

hms-app/src/pages/IdeationPortal.tsx
LetsPoke: thats to find the one hackathon with "null date"

hms-app/src/common/redux/hackathonSlice.ts
```diff
@@ -11,7 +11,8 @@ const initialState: HackathonState = {
   hackathons: [
     {
       id: '33a892ba-8a9e-4edf-947a-f421c86a14b5',
-      title: '',
+      title: 'title',
+      slug: 'slug',
```
verzano, line 15: slugs must be unique :P

hms-app/src/common/redux/hackathonSlice.ts
```diff
@@ -11,7 +11,8 @@ const initialState: HackathonState = {
   hackathons: [
```
verzano, line 11: if this is the list of all hackathons, shouldn't it contain the next and previous ones as well?

hms-app/src/common/constants.ts
verzano: and since the IdeationPortal hackathon exists in the backend it cannot have a null date 🤦

ok yeah I get it all now, the `MIN_DATE` is for sure a hack but makes sense

hms-app/src/components/HackathonSelectDropdown.tsx
verzano: ok, it's just unclear as to what this does at first glance.  I get it all now though.  There are better ways we could do this but for now we can leave it

hms-app/src/App.tsx
```diff
@@ -26,6 +26,7 @@ function App() {
       <Route path='/' element={<Layout />}>
         <Route index element={<Home />} />
         <Route path='ideas' element={<AllIdeas />} />
+        <Route path='ideas/:slug' element={<AllIdeas />} />
```
verzano, line 29: I think this is the more common path.  I wonder though if it should be like `idea/slug/:slug` or just `slug/:slug`.  That way we could still have `idea/:idea_id` 🤔 

```suggestion
        <Route path='idea/:slug' element={<AllIdeas />} />
```

verzano: I also call this "more common" based mostly on how github does `...github.com.../pulls` and `...github.com.../pull/1252`

hms-app/src/common/constants.ts
LetsPoke: yeah bit hacky but for now the best ways I would say

hms-app/src/common/redux/hackathonSlice.ts
LetsPoke: its just to show redux how a hackathon looks like

hms-app/src/App.tsx
LetsPoke: we shouldnt use the word slug in the url as it could confuse people but maybe call it hackathons/:slug to make it easier to understand 

Maybe if we decide to have own pages for the idea details we could use something like idea/:ideaId

PR #1268, opened 2023-02-08
Fix: Rich Text Editor not working

### LINKED ISSUE #806: rich texteditor is broken since version update
rich text editor is not usable and displayable on different pages since version update
need to investigate and solve

LetsPoke:
fixed the not working RTE

closes #806

### DIFF
--- a/hms-app/src/components/card-details/HackathonDetails.tsx
+++ b/hms-app/src/components/card-details/HackathonDetails.tsx
@@ -61,16 +61,11 @@ export default function HackathonDetails(props: IProps) {
   const [relevantIdeaList, setRelevantIdeaList] = useState([] as Idea[])
   const [votingOpened, setVotingOpened] = useState<boolean>(false)
 
-  const initialValue =
-    '<p>Please add your <b>hackathon description</b> here</p>'
-  const [descriptionValue, onChange] = useState(initialValue)
-
   const loadSelectedHackathon = () => {
     getHackathonDetails(instance, hackathonId).then(
       (data) => {
         setHackathonData(data)
         setVotingOpened(data.votingOpened)
-        onChange(data.description || '')
         setIsHackathonLoading(false)
         setIsHackathonError(false)
       },
@@ -303,8 +298,7 @@ export default function HackathonDetails(props: IProps) {
                 <RichTextEditor
                   readOnly
                   value={hackathonData.description || ''}
-                  onChange={onChange}
-                  id='rte'
+                  id='hackathonDescriptionEditor'
                 />
               </Container>
             )}
@@ -356,8 +350,7 @@ export default function HackathonDetails(props: IProps) {
               <RichTextEditor
                 readOnly
                 value={hackathonData.description || ''}
-                onChange={onChange}
-                id='rte'
+                id='hackathonDescriptionEditor'
               />
             </Card.Section>
 
--- a/hms-app/src/components/card-details/HackathonDetails.tsx
+++ b/hms-app/src/components/card-details/HackathonDetails.tsx
@@ -30,7 +30,6 @@ import HackathonForm from '../input-forms/HackathonForm'
 import CategoryDetails from './CategoryDetails'
 import { Link } from 'react-router-dom'
 import { styles } from '../../common/styles'
-import { RichTextEditor } from '@mantine/rte'
 import { NULL_DATE } from '../../common/constants'
 import HackathonHeader from '../HackathonHeader'
 import { useMsal } from '@azure/msal-react'
@@ -41,6 +40,7 @@ import {
 } from '../../common/colors'
 import { showNotification, updateNotification } from '@mantine/notifications'
 import { Check, X } from 'tabler-icons-react'
+import { RichTextEditor } from '@mantine/rte'
 
 type IProps = {
   hackathonId: string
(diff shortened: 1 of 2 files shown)


hms-app/src/components/card-details/HackathonDetails.tsx
```diff
@@ -59,15 +59,18 @@ export default function HackathonDetails(props: IProps) {
   const [hackathonData, setHackathonData] = useState({} as Hackathon)
   const [ideaData, setIdeaData] = useState<Idea>()
   const [relevantIdeaList, setRelevantIdeaList] = useState([] as Idea[])
-  const [value, onChange] = useState(hackathonData.description)
   const [votingOpened, setVotingOpened] = useState<boolean>(false)
 
+  const initialValue =
```
verzano, line 64: naming it something like `descriptionHint` makes it a bit more obvious to it's purpose

hms-app/src/components/input-forms/HackathonForm.tsx
```diff
@@ -207,9 +210,9 @@ function HackathonForm(props: IProps) {
             Description
           </Title>
           <RichTextEditor
-            value={DescriptionValue}
+            value={descriptionValue}
             onChange={onChange}
-            style={{ minHeight: 225 }}
+            id='rte'
```
verzano, line 215: the ID could be a good bit more descriptive (and unique), `hackathonDescriptionEditor` or so

hms-app/src/components/input-forms/HackathonForm.tsx
```diff
@@ -25,9 +25,12 @@ function HackathonForm(props: IProps) {
   const [endDateValue, setEndDateValue] = useState<Date | null>(new Date())
   const [hackathonTitle, setHackathonTitle] = useState('')
   const [hackathonSlug, setHackathonSlug] = useState('')
-  const [DescriptionValue, onChange] = useState('')
   const [hackathon, setHackathon] = useState<Hackathon>({} as Hackathon)
 
+  const initialValue =
+    '<p>Please add your <b>hackathon description</b> here</p>'
+  const [descriptionValue, onChange] = useState(initialValue)
```
verzano, line 32: I don't think `onChange` is a good name for this.  Basically react creates a function with that name under the hood, a function named simply `onChange` doesn't provide enough context as to what it is doing.  `setDescriptionValue` I think would be the best practice here

hms-app/src/components/input-forms/HackathonForm.tsx
```diff
@@ -25,9 +25,12 @@ function HackathonForm(props: IProps) {
   const [endDateValue, setEndDateValue] = useState<Date | null>(new Date())
   const [hackathonTitle, setHackathonTitle] = useState('')
   const [hackathonSlug, setHackathonSlug] = useState('')
-  const [DescriptionValue, onChange] = useState('')
   const [hackathon, setHackathon] = useState<Hackathon>({} as Hackathon)
 
+  const initialValue =
```
verzano, line 30: a better name for this would be good something like `hackathonDescriptionHint`

hms-app/src/components/card-details/HackathonDetails.tsx
```diff
@@ -59,15 +59,18 @@ export default function HackathonDetails(props: IProps) {
   const [hackathonData, setHackathonData] = useState({} as Hackathon)
```
verzano, line 59: so I'm guessing this whole issues is stemming from this line here.  If you were to give description a value here it would fix it, so...
```suggestion
  const [hackathonData, setHackathonData] = useState({
    description: ''
  } as Hackathon)
```

In general all the places where you have been using these types might benefit from them being classes instead.  And then you could have default constructors that make a good "empty" instance of the classes.

hms-app/src/components/card-details/HackathonDetails.tsx
```diff
@@ -297,9 +300,12 @@ export default function HackathonDetails(props: IProps) {
...
               <Container mb={25}>
-                <RichTextEditor readOnly value={value!} onChange={onChange}>
-                  {hackathonData.description}
-                </RichTextEditor>
+                <RichTextEditor
+                  readOnly
+                  value={hackathonData.description || ''}
+                  onChange={onChange}
+                  id='rte'
+                />
```
verzano, line 308: this should probably use another option as well with your `initialValue` constant, something like.

```suggestion
                <RichTextEditor
                  readOnly
                  value={hackathonData.description || ''}
                  onChange={onChange}
                  placeholder={intialValue}
                  id='rte'
                />
```

I'm not sure if this extension is part of mantine's RichTextEditor by default, but I saw examples of it's use in their docs

hms-app/src/components/card-details/HackathonDetails.tsx
```diff
@@ -59,15 +59,18 @@ export default function HackathonDetails(props: IProps) {
   const [hackathonData, setHackathonData] = useState({} as Hackathon)
   const [ideaData, setIdeaData] = useState<Idea>()
   const [relevantIdeaList, setRelevantIdeaList] = useState([] as Idea[])
-  const [value, onChange] = useState(hackathonData.description)
   const [votingOpened, setVotingOpened] = useState<boolean>(false)
 
+  const initialValue =
+    '<p>Please add your <b>hackathon description</b> here</p>'
+  const [descriptionValue, onChange] = useState(initialValue)
```
verzano, line 66: it actually looks like you never use this value anywhere 🤔

you set it whenever someone types in the RichTextEditor but I don't see it being access anywhere

verzano: ah, because it is read only... ok... then

I don't think you need this tbh, you should really just be getting the value from the hackathon which is  provided to this class.  In addition you pass in a `onChange` listener to the RichTextEditor which it probably doesn't need at all (since it is not editable)

The backend also is meant to ensure that a description is never null so you shouldn't actually need the checks that you have everywhere (`xxx || ''`), you should just construct a "valid" hackathon in the place where you make the state hook

hms-app/src/components/card-details/HackathonDetails.tsx
verzano: so this should apply to the `HackathonForm.tsx` file and not this one since you should never need a Placeholder because there should be a value here or nothing

hms-app/src/components/input-forms/HackathonForm.tsx
verzano: if you change it to a use a placeholder, then you could just in-line the value there and don't bother with a constant

verzano: also with text editors in general allowing HTML can be... scary to say the least.  My personal view on the matter is to permit some form of markdown as it allows for nice pretty formatted things, but not so much freedom that you can build a scam website in a form and possibly trick people into doing something dumb

LetsPoke: The RTE is only accessible for admins so I think this would be ok 
Admins as we have them could also directly manipulate the source code so no need for hacky html tricks

PR #1277, opened 2023-02-27
Mock resolved value once

LetsPoke:
changed mockResolvedValue into mockResolvedValueOnce everywhere 

for now manually reset mock data where it is needed 
will be touched within the next Ticket

closes #1271

PR #1279, opened 2023-02-27
1278 switch from clearmock to resetmock

### LINKED ISSUE #1278: switch from clearMock to resetMock
we want to switch from clearMock to resetMock in the jest.config to ensure that we work with correct mocked test data 
(because clearMock does not work with "spyOn")

LetsPoke:
switched to resetMock 
and fixed all related tests 

only merge after #1277 

closes #1278


.github/workflows/backend-tests.yml
```diff
@@ -1,6 +1,10 @@
 name: Backend (HMS API) Tests
 
-on: pull_request
+on:
+  pull_request:
+  push:
+    branches:
+      - '*'
```
verzano, line 7: so I was attempting to avoid it doing too many runs in case you did a lot of pushes, but I doubt we would exceed some limit in GitHub to be honest

verzano:
was this one originally branched from main or no?

verzano (approved):

PR #1303, opened 2023-03-06
1276 increase code coverage

LetsPoke:
added and updated a lot of tests 

only merge after #1277 and #1279


hms-api/src/service/idea-comment-service.ts
```diff
@@ -32,12 +30,6 @@ export async function createIdeaComment(
     throw new ReferenceNotFoundError(`User with id: ${userId} not found`);
   }
 
-  if (parentCommentId && !(await parentCommentIdIdeaExists(parentCommentId))) {
-    throw new ReferenceNotFoundError(
-      `Parent comment with id: ${parentCommentId} not found`,
-    );
-  }
```
verzano, line 39: why is this removed?

hms-api/src/handler/ideaComment/create.ts
```diff
@@ -7,10 +7,9 @@ import IdeaCommentCreateResponse from '../../rest/ideaComment/IdeaCommentCreateR
 export async function create(event, context, callback) {
   await wrapHandler(async () => {
     const request = IdeaCommentCreateRequest.parse(event.body);
-    const ideaId = event.pathParameters.id;
     const ideaComment = await createIdeaComment(
       request.userId,
-      ideaId,
```
verzano, line 13: idk about this 🤔

hms-api/src/error/ValidationError.ts
```diff
@@ -4,7 +4,7 @@ import ValidationResult from './ValidationResult';
 
 class ValidationError extends Error {
   constructor(message: string, result: ValidationResult) {
-    super(message + ' ' + result.toBulletList());
```
verzano, line 7: was it making double spaces?

LetsPoke: yes it always added one extra space to the "message"

hms-api/src/service/idea-comment-service.ts
LetsPoke: for consistency as we nowwhere else check for partencommentId neither test for it 
just to have it completly removed for now until we someday add partenCommentIds

hms-api/src/handler/ideaComment/create.ts
LetsPoke: im sure there was a good reason why i did this 
maybe it was just for consistency with all the other create methods

hms-api/src/service/idea-comment-service.ts
verzano: then I would argue that you should either leave it (so that you don't have to remember later to put it back) or you remove everything related to parent comments

hms-api/src/handler/ideaComment/create.ts
verzano: well, we SHOULD be pulling it from the path and not from the request

verzano: or, we should remove it from the path (means changing the frontend too) and only get it from the request

hms-api/src/service/idea-comment-service.ts
LetsPoke: just think about as it was never there

hms-api/src/handler/ideaComment/create.ts
LetsPoke: if this dont make the tests impossible then i can try it

hms-api/src/service/idea-comment-service.ts
verzano: but, it IS there.  In the function arguments.  There is no reason to not just leave this here

Issue #1335, opened 2023-04-03
fix: I cant edit or move ideas in the "my ideas" page


LetsPoke:

rz-p:
Quick question regarding `IdeaDetails.tsx` @LetsPoke 

Does every idea card has an assigned default type or it can be assigned multiple types?

LetsPoke:
yeah its always 1 Type 
and if I remember correctly the type defines the context in which the idea is shown
for example if the idea is loaded in the ideation portal then the type is set to ideation portal

rz-p:
Does the type also cascades? I see there are many `&&` and `||` arguments and when I changed some of it, elements from other types also shows up

PR #1337, opened 2023-04-12
#1335 Fix: i cant edit or move ideas in the my ideas page

rz-p:
Changed the argument to && if either `Type.Admin` or `Type.Owner` is true

### DIFF
--- a/hms-app/src/components/card-details/IdeaDetails.tsx
+++ b/hms-app/src/components/card-details/IdeaDetails.tsx
@@ -661,7 +661,7 @@ export default function IdeaDetails(props: IProps) {
                     )}
 
                     {type === IdeaCardType.Admin ||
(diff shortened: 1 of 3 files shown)


hms-app/src/components/card-details/IdeaDetails.tsx
LetsPoke, line 1: is it intended that you cant edit the ideas in the "hackathons" page now?

rz-p: You actually can (at least on my local instance)

rz-p:
Idea owner specific button is now rendered on both `All Ideas` and `My Ideas` page. Not sure if it can be seen by other users though.

LetsPoke (approved):

Issue #1363, opened 2023-04-28
fix: refresh in admin page 


reyhanrumengan:
1. Create New Hackathon:
After create new hackathon, new hackathon can be seen in the list and form should be cleared

2. Hackathon List:
- After delete, list should refresh
- After edit, modal should close automatically and info should refresh.
-  After pressing reload, info should refresh without reloading page (now: only the hackathonDetails are reloaded (if change Title, the Title in the Hackathon List stays the same). **Question: Why put reload button here ? What use?**

3. Hackathon List -> Category:
- After delete, list should refresh without reloading page
- After edit, info should refresh without reloading page

4. Hackathon List -> Participant:
- After delete, list should refresh without reloading page

5. Hackathon List -> Ideas: (the button already there in other branches / PR)
- After delete, list should refresh without reloading page
- After edit, info should refresh without reloading page

6. User List:
- After delete, user should be deleted and list should refresh without reloading page
- After edit, modal should close automatically and info should refresh without reloading page
**- Do reload button useful here?**

PR #1401, opened 2023-05-03
#1363 fix: refresh in admin page -> hackathonlist

reyhanrumengan:


LetsPoke (approved):

hms-app/src/components/card-details/HackathonDetails.tsx
```diff
@@ -197,6 +202,10 @@ export default function HackathonDetails(props: IProps) {
     </Modal>
   )
 
+  const closeEditModal = (isOpened: boolean) => {
+    setEditModalOpened(isOpened)
+  }
```
verzano, line 207: the name of this function doesn't really match what it does, in fact, why even have this function at all when you just pass the value right into another function?

hms-app/src/components/input-forms/HackathonForm.tsx
```diff
@@ -18,11 +18,13 @@ type IProps = {
   context: string
   hackathonId: string | null
   closeAccordion?: () => void
+  reload?: () => void
+  setOpened?: (boolean: boolean) => void
```
verzano, line 22: boolean is a VERY bad name for a variable (in fact it is probably should not even be allowed)

verzano: I'm also curios as to why all of these callbacks are nullable, they sound somewhat important and I wonder if they ever are/should be null

verzano: setOpened is also a really generic name for a function and doesn't well describe what it does.  A good rule of thumb for callbacks is to name the for the thing which happened (things like; `onUserDeleted`, `beforeTaskShutdown`, or `afterBombExploded`).

reyhanrumengan: > I'm also curios as to why all of these callbacks are nullable, they sound somewhat important and I wonder if they ever are/should be null

It should be nullable, because in Admin, we only use the closeAccordion. 
And in HackathonDetails, we only use the reload and setOpened
HackathonForm is being used in 2 different places with different argument that are being passed

PR #1403, opened 2023-05-03
fix: naming convention

reyhanrumengan:


reyhanrumengan:
@LetsPoke 
Should I also change the naming convention from IdeaForm, IdeaDetails, etc (idea related stuffs)? Or maybe do it later?
Because I think currently Rizky and Katja are working on the idea related stuffs. 
Just so that there won't be a lot of conflict happening

hms-app/src/components/input-forms/HackathonForm.tsx
```diff
@@ -19,12 +19,12 @@ type IProps = {
...
   closeAccordion?: () => void
   reload?: () => void
-  setOpened?: (boolean: boolean) => void
+  closeEditModal?: () => void
 }
 
 function HackathonForm(props: IProps) {
   const { instance } = useMsal()
-  const { context, hackathonId, closeAccordion, reload, setOpened } = props
+  const { context, hackathonId, closeAccordion, reload, closeEditModal } = props
```
verzano, line 27: ok so having looked at this `closeAccordion` and `closeEditModal` are used when the same thing happens.  Both are called when operation that is meant to be performed is completed successfully.

as well `reload` is always called with `closeEditModal` so the two could be combined into a single thing.

I would say that you replace all three callbacks with a single one called `onOperationSuccess`

LetsPoke: this sounds really good

LetsPoke:
> @LetsPoke Should I also change the naming convention from IdeaForm, IdeaDetails, etc (idea related stuffs)? Or maybe do it later? Because I think currently Rizky and Katja are working on the idea related stuffs. Just so that there won't be a lot of conflict happening

Maybe we can open an issue for renaming these things and tackle it next week after the changes from rizky and katja are done

hms-app/src/components/input-forms/HackathonForm.tsx
reyhanrumengan: and now it became simpler to look into 👍

hms-app/src/components/HeaderMenu.tsx
```diff
@@ -227,8 +227,9 @@ export default function HeaderMenu({
     </div>
   )
 
-  const closeEditModal = (isOpened: boolean) => {
-    setEditModalOpened(isOpened)
+  const closeEditModal = () => {
```
LetsPoke, line 230: now that name doesnt fit anymore

reyhanrumengan: is it because it also refresh the list? So the function name doesn't fit

LetsPoke: yeah thats what I mean 
its not closing the edit modal as the name suggest, its more like refreshAfterChange or somethink like that

hms-app/src/components/card-details/HackathonDetails.tsx
```diff
@@ -218,13 +211,12 @@ export default function HackathonDetails(props: IProps) {
       <HackathonForm
         context={'edit'}
         hackathonId={hackathonData.id}
-        reload={refreshList}
-        setOpened={closeEditModal}
+        onSuccess={refreshAfterChange}
       />
       {isHackathonLoading && <div>Loading...</div>}
       <Text className={classes.text}>
         (This window will automatically closed as soon as the hackathon is
```
verzano, line 218:
```suggestion
        (This window will automatically close after the hackathon is
```

hms-app/src/components/card-details/UserDetails.tsx
```diff
@@ -17,7 +17,7 @@ import { DELETE_BUTTON_COLOR, JOIN_BUTTON_COLOR } from '../../common/colors'
 
 type IProps = {
   userId: string
-  reloadList?: () => void
+  reloadList: () => void
```
verzano, line 20: this could probably be renamed to `onUserDeleted` (that seems to be the only action that causes the callback to be triggered)

verzano: if, however, it is used elsewhere then some other name would be appropriate.  The issue with `reloadList` as a name is that from within this file you don't know what "list" is, and list might not make any sense to a new component which will make use of the `UserDetails`

hms-app/src/components/card-details/HackathonDetails.tsx
```diff
@@ -44,7 +44,7 @@ import { RichTextEditor } from '@mantine/rte'
 type IProps = {
   hackathonId: string
   type: HackathonDetailsType
-  reloadList?: () => void
+  reloadList: () => void
```
verzano, line 47: same thing about the name here, seems to only be used when a hackathon is deleted

reyhanrumengan: here, it also used for when a hackathon is edited. 
What proper name to used for change and delete?
`onChanged`(?)

verzano: `onSuccess` seems good enough

verzano (approved):

Issue #1350, opened 2023-04-24
fix: ideas in the ideation portal should disappear by itself when delete


katty-bird:
acceptance criteria:

when I delete an idea in the ideation portal the idea card should disappear on its own from the list of ideas

PR #1406, opened 2023-05-05
#1350 update idea list in the ideation portal when delete idea

katty-bird:
added updating of the ideas list in the ideation portal when an idea is deleted

### DIFF
--- a/hms-app/src/components/card-details/CardButton.tsx
+++ b/hms-app/src/components/card-details/CardButton.tsx
@@ -14,6 +14,7 @@ import MoveIdeaModal from '../MoveIdeaModal'
 type IProps = {
   idea: Idea
   reloadIdeaDetails?: () => void
+  reloadIdeaList: () => void
 }
 
 export default function CardButtons(props: IProps) {
@@ -22,7 +23,7 @@ export default function CardButtons(props: IProps) {
 
   const { instance } = useMsal()
   const { classes } = styles()
-  const { idea, reloadIdeaDetails } = props
+  const { idea, reloadIdeaDetails, reloadIdeaList } = props
   const [ideaData, setIdeaData] = useState(idea)
   const [loader, setLoader] = useState(false)
 
@@ -37,6 +38,7 @@ export default function CardButtons(props: IProps) {
     })
     deleteIdea(instance, ideaData.id).then((response) => {
       setDeleteModalOpened(false)
+      reloadIdeaList()
       if (JSON.stringify(response).toString().includes('error')) {
         updateNotification({
           id: 'delete-idea-load',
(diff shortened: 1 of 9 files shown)


hms-app/src/components/card-details/IdeaDetails.tsx
```diff
@@ -39,6 +39,7 @@ type IProps = {
   idea: Idea
   isLoading: boolean
   type: IdeaCardType
+  reloadIdeaList: () => void
```
LetsPoke, line 42: if this is not nullable we have to add this in all files that use "IdeaDetails"

hms-app/src/components/card-details/CardButton.tsx
```diff
@@ -14,6 +14,7 @@ import MoveIdeaModal from '../MoveIdeaModal'
 type IProps = {
   idea: Idea
   reloadIdeaDetails?: () => void
+  reloadIdeaList: () => void
```
LetsPoke, line 17: should also be nullable here then

hms-app/src/components/lists/IdeaCardList.tsx
```diff
@@ -8,16 +8,17 @@ type IProps = {
   isLoading: boolean
   columnSize: number
   type: IdeaCardType
+  reloadIdeaList: () => void
```
LetsPoke, line 11: same here

hms-app/src/components/card-details/CardButton.tsx
verzano: these seem like a pretty similar scenario to the previous PR I was looking at that where these could be named a bit differently, here I would say they could be `onIdeaEdited` and `onIdeaDeleted`.

This whole file could/should also probably be named differently, it is called `CardButton` but appears to be specific to an Idea and is likely something more like `IdeaCardButtonGroup` or so

verzano: I would even be tempted to say this should be split into two things, a `DeleteIdeaButton` and an `EditIdeaButton` and then instead of having this `CardButtons`, just using those to components directly in the `IdeaDetails` 🤷

verzano: but if you always use the edit and delete buttons as a pair then this makes sense as it is

LetsPoke: 1. yeah similar case and we already opened a ticket for the button/idea files but will only tackle them after the refactoring is done so that we dont create more conflicts then nessesary
2. splitting it even more would be not a bad idea but yeah right now we always use these buttons in combination

LetsPoke (approved):

PR #1419, opened 2023-05-10
fix: refresh Hackathons/All Ideas page after deleting an idea

katty-bird:
Added reloadHackathon function that changes hackathon ID to make RelevantIdeasLoader work  -> All ideas page updates itself after an idea is deleted

### DIFF
--- a/hms-app/src/pages/AllIdeas.tsx
+++ b/hms-app/src/pages/AllIdeas.tsx
@@ -77,6 +77,12 @@ function AllIdeas() {
     return item.title?.toLowerCase().includes(searchTerm.toLowerCase())
   })
 
+  const reloadHackathon = () => {
+    const id = selectedHackathonId 
+    setSelectedHackathonId('')
+    setSelectedHackathonId(id)
+  }
+
   const findParticipant = () => {
     let participant: ParticipantPreview
     if (hackathonData.participants) {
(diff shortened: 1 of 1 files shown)


hms-app/src/pages/AllIdeas.tsx
```diff
@@ -77,6 +77,12 @@ function AllIdeas() {
     return item.title?.toLowerCase().includes(searchTerm.toLowerCase())
   })
 
+  const reloadHackathon = () => {
+    const id = selectedHackathonId 
+    setSelectedHackathonId('')
+    setSelectedHackathonId(id)
```
verzano, line 83: this feels really hacky... I would argue it is likely not correct

verzano: I wonder why we also have both `selectedHackathonId` state and `hackathonData` state (which has it's own ID).  Seems like just having `hackathonData` would be enough

verzano: what you really wanna be doing is to update the `relevantIdeasList` I would think

verzano: the reason I think this works right now (and not before) is because you end up changing the value of the selected hacakthon ID (before you were setting it to the same thing so nothing changed)

LetsPoke: The main reason for this problem is the "RelevantIdeasLoader" that is part of the old pages (all ideas, my ideas and archive) I would say that we should refactor this whole thing so that we can simplify everything
and your 100% right that is somewhat hacky

verzano: beyond that this is just a bit wrong, the thing that actually changed is an item in the `relevantIdeasList` and the set function for that should be used.  You're right though to fix it you really need to get rid of the `RelevantIdeasLoader` and instead probably just use the calls to get one or more ideas directly to refresh your list.

PR #1421, opened 2023-05-10
fix: skills are not loading and not displayed when edit user

reyhanrumengan:


reyhanrumengan:
@LetsPoke does it make sense to pass the `userSkill` from the `userDetails` for the checkbox `defaultValue`?

LetsPoke (approved):

PR #1423, opened 2023-05-11
refactor: pass user instead user id to editUserForm

reyhanrumengan:


hms-app/src/components/input-forms/EditUserForm.tsx
```diff
@@ -32,28 +31,20 @@ export default function EditUserForm(props: IProps) {
...
-    )
+    setIsLoading(false)
+    setUserData({
+      id: user.id,
+      lastName: user.lastName as any,
+      firstName: user.firstName,
+      emailAddress: user.emailAddress,
+      roles: user.roles as any,
+      imageUrl: user.imageUrl as any,
+    })
```
reyhanrumengan, line 42: Does it make sense to set the `userData` after getting the `user` from the `UserDetails`/`HeaderMenu` ?

Because to edit a user, we already have `editUser` function in `userAction`. And this function only works with the specific properties like in the userData that we set

reyhanrumengan, line 42: Does it make sense to set the `userData` after getting the `user` from the `UserDetails`/`HeaderMenu` ?

Because to edit a user, we already have `editUser` function in `userAction`. And this function only works with the specific properties like in the userData that we set

hms-app/src/components/input-forms/EditUserForm.tsx
```diff
@@ -66,7 +57,7 @@ export default function EditUserForm(props: IProps) {
       autoClose: false,
       disallowClose: false,
     })
-    editUser(instance, user, skills).then((response) => {
+    editUser(instance, userData, skills).then((response) => {
```
reyhanrumengan, line 60: This is the `editUser` part that I mentioned before.

hms-app/src/components/input-forms/EditUserForm.tsx
LetsPoke: yes we dont need the getUserDetails function anymore when we pass down the whole user

hms-app/src/components/input-forms/EditUserForm.tsx
```diff
@@ -1,59 +1,37 @@
...
+    firstName: user.firstName,
+    emailAddress: user.emailAddress,
+    roles: user.roles,
+    imageUrl: user.imageUrl ? user.imageUrl : '',
   }
 
   useEffect(() => {
-    loadSelectedUser()
     loadAvailableSkills()
-    setIsLoading(true)
```
LetsPoke, line 56: this can stay here if we set it false in the load skill functions

LetsPoke (approved):

PR #1426, opened 2023-05-12
fix: refresh ideation portal after idea is moved

katty-bird:
added the idea list update on the Ideation portal when an idea is moved to the hackathon page

### DIFF
--- a/hms-app/src/components/MoveIdeaModal.tsx
+++ b/hms-app/src/components/MoveIdeaModal.tsx
@@ -12,9 +12,12 @@ import { removeIdeaParticipant } from '../actions/ParticipantActions'
 
 type IProps = {
   idea: Idea
+  reloadIdeaList?: () => void
 }
 
-export default function MoveIdeaModal({ idea }: IProps) {
+export default function MoveIdeaModal(props: IProps) {
+  
+  const { idea, reloadIdeaList } = props
   const { instance } = useMsal()
   const { classes } = styles()
   const [opened, setOpened] = useState(false)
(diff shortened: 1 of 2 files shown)


hms-app/src/components/MoveIdeaModal.tsx
```diff
@@ -12,9 +12,12 @@ import { removeIdeaParticipant } from '../actions/ParticipantActions'
 
 type IProps = {
   idea: Idea
+  reloadIdeaList?: () => void
```
LetsPoke, line 15: can we rename this into "onSuccess"

hms-app/src/components/card-details/CardButton.tsx
```diff
@@ -145,7 +145,10 @@ export default function CardButtons(props: IProps) {
         Edit
       </Button>
       <FinalVideoUploadModal idea={ideaData} />
-      <MoveIdeaModal idea={ideaData} />
+      <MoveIdeaModal 
+        idea={ideaData}
+        reloadIdeaList={reloadIdeaList}
```
LetsPoke, line 150: and rename this into "onSuccess={reloadIdeaList}"

hms-app/src/components/MoveIdeaModal.tsx
LetsPoke: oh and it doesnt need the "?" as we only call MoveIdeaModal only once and when then we pass that for sure

hms-app/src/components/MoveIdeaModal.tsx
```diff
@@ -71,6 +74,7 @@ export default function MoveIdeaModal({ idea }: IProps) {
         if (setOpened) {
           setOpened(false)
         }
+        if (reloadIdeaList) reloadIdeaList()
```
LetsPoke, line 77: if we remove the "?" then we dont need the if statement here anmyore

LetsPoke (approved):

PR #1437, opened 2023-05-31
fix: updating page after idea creation

reyhanrumengan:


hms-app/src/pages/AllIdeas.tsx
```diff
@@ -57,6 +57,7 @@ function AllIdeas() {
     if (hackathonData.id !== undefined) {
       setOpened(false)
     }
+    reloadHackathon()
```
verzano, line 60: so what exactly should reload in this component? sure it's the "hackathon" but what data might have been changed when the modal is closed?

reyhanrumengan: it reloads the hackathon after an Idea were being deleted.
maybe its better using `reloadAfterDeleted` ?

reyhanrumengan: it should reloads the hackathon after the idea were edited / deleted

verzano: ok so it is the `IdeaCardsList` that is refreshed either by removing the deleted idea or updating the edited one.  I remember @LetsPoke saying that the idea was to redo that `RelevantIdeasLoader` to do this better.  So I suppose it is fine for now

verzano (approved):

hms-app/src/pages/AllIdeas.tsx
LetsPoke: yeah we have a open ticket to change the whole logic and everything that has to do with loading thing because this RelevantIdeasLoader is garbage

LetsPoke (approved):

PR #1438, opened 2023-05-31
fix: rename ideation portal page -> idea pool

reyhanrumengan:


hms-app/src/components/HeroHeader.tsx
```diff
@@ -185,7 +185,7 @@ const HeroHeader = (props: { nextHackathon: HackathonSerializable }) => {
         <div className={classes.controls}>
           <Button
             component={Link}
-            to='/ideas'
+            to='/hackathons'
```
verzano, line 188: does this make sense? the button text makes it sound like it would go to the idea pool page

LetsPoke (approved):

hms-app/src/components/HeroHeader.tsx
verzano: hmm still confusing, seems like this button would take me to the list of ALL hackathons, and not the next hackathon's ideas 🤔

LetsPoke: Bildschirm­foto 2023-05-31 um 15 49 14

as you can see here the "all ideas" page in the navbar has the link "hackathons" 
open for name suggestions

LetsPoke: maybe hackathons/all-ideas make sense but we dont have a root page for hackathons then

verzano: honestly those don't sound like connected things a menu heading called "hackathons" I would think would contain things like "my hackathons", "all hackathons", "past hackathons" or would just take you to a page with all the hackathons and not have a submenu.

"voting" sounds like it should be a sub-item of only 1 hackathon and not many hackathons

"all ideas" and "my ideas" sound like they should be sub-items of the "idea pool" or just "ideas"

verzano: personally I would think "home" would be either a list of hackathons or the one you are currently competing in, then you'd have "ideas" and "hackathons" which would have filters on the pages to present you with less or more of those things as you want

LetsPoke: that sounds like a complete different concept of the whole app :) 
legit points but complete different structure of the subpages and the content presentation

would be nice to have a whole concept talk in the future (also with more then just us 2 )

verzano: agreed, lets try and set something up

PR #1465, opened 2023-06-05
fix: in edit mode the edit button renamed to save

katty-bird:


reyhanrumengan:
@katty-bird Can you also change the edit button for the other forms? Like the `CategoryForm`, `EditUserForm`,  etc

LetsPoke (approved):

PR #1470, opened 2023-06-05
gap between light/dark switch and home is increased

katty-bird:


hms-app/src/common/styles.ts
```diff
@@ -74,7 +74,7 @@ export const styles = createStyles((theme) => ({
   link: {
     display: 'block',
     lineHeight: 1,
-    padding: '8px 12px',
+    padding: '6px 12px',
```
LetsPoke, line 77: pls revert his change 
I think its unrelated to the toggle

LetsPoke (approved):

PR #1477, opened 2023-06-06
fix: upload presentation only after hackathon start

reyhanrumengan:


reyhanrumengan:
@LetsPoke currently i didn't made the changes for the myIdeas page, because I thought we discussed that we want to remove this page. What are your thought? Should I for now also implement this in the myIdeas page?

hms-app/src/components/buttons/CardButton.tsx
```diff
@@ -122,6 +124,13 @@ export default function CardButtons(props: IProps) {
     </Modal>
   )
 
+  const uploadButton = () => {
+    return (
+      type === IdeaCardType.AllIdeas &&
```
LetsPoke, line 129: do we really need that check? 
I would say we would need to check if its the admin page or not
on all pages it should only be visible when hackathon started, excpect for the admin page there it should be always visible

LetsPoke:
> @LetsPoke currently i didn't made the changes for the myIdeas page, because I thought we discussed that we want to remove this page. What are your thought? Should I for now also implement this in the myIdeas page?

as it should be not much of work I would also add it to myideas and ideaPool

hms-app/src/components/buttons/CardButton.tsx
reyhanrumengan: that does makes sense. I forgot about the admin page

LetsPoke (approved):

PR #1486, opened 2023-06-07
feat: add idea creation date

reyhanrumengan:


hms-app/src/components/card-details/IdeaDetails.tsx
```diff
@@ -296,6 +304,7 @@ export default function IdeaDetails(props: IProps) {
             </Accordion.Item>
           </Accordion>
           {IdeaComments()}
+          {ideaCreationDate()}
```
LetsPoke, line 307: can we put it below the idea desribtion

hms-app/src/components/card-details/IdeaDetails.tsx
```diff
@@ -253,6 +253,14 @@ export default function IdeaDetails(props: IProps) {
     }
   }
 
+  const ideaCreationDate = () => {
+    return (
+      <Text className={classes.smallText}>
+        {new Date(idea.creationDate).toDateString()}
```
LetsPoke, line 259: can we add an "Creation Date: " before the actual date so that is 100% clear to the user what this date is

hms-app/src/components/card-details/IdeaDetails.tsx
```diff
@@ -278,6 +278,7 @@ export default function IdeaDetails(props: IProps) {
               {ideaDescription()}
             </Card.Section>
           </Spoiler>
+          {ideaCreationDate()}
```
LetsPoke, line 281: inside the spoiler was what i meant :)

LetsPoke: so directly below the description (inside spoiler and section)

reyhanrumengan: if the description is long, then the date wont be shown unless we click "show more" (?)

reyhanrumengan: Bildschirmfoto 2023-06-07 um 08 47 38
it will be like that

LetsPoke: yeah thats what I mean 
as not everybody may be interested in the creation date is it is no crucial information we can but in there so that people can find it when they are interested in the whole idea but will not directly see it on the card 
(maybe we could also put it somewhere into the "show details" accordion

LetsPoke (approved):

PR #1503, opened 2023-06-12
fix: save last selected hackathon for the dropdown

reyhanrumengan:


hms-app/src/components/HackathonHeader.tsx
```diff
@@ -35,26 +35,6 @@ export default function HackathonHeader(props: IProps) {
             />
           </Center>
         )}
-        <Center mt={10}>
```
reyhanrumengan, line 38: @LetsPoke Do we want to remove the date badge on the `HackathonHeader`? Because right now, when there is no hackathon that being selected, the badge still shows (showing today's date)

LetsPoke: lets remove this next time when we have a new version

Issue #1496, opened 2023-06-09
BA: Update IdeaForm


LetsPoke:
In the Ideaform -> Skills and Category section is confusing 

We should rename "Required Skills" into "Usefull Skills" and add a description that people COULD choose 1 or more skills that they think are usefull for the ideas 

We should add a descrption to the category, explaining that it means hackathon related categories 
and update description that they can choose 1 not 1 or multiple

PR #1506, opened 2023-06-12
#1496 ba: descriptions for skills and categories changed

katty-bird:


hms-app/src/components/input-forms/IdeaForm.tsx
```diff
@@ -300,8 +300,8 @@ function IdeaForm(props: IProps) {
 
             <Card.Section className={classes.borderSection}>
               <Checkbox.Group
-                label='Required skills'
-                description='chose one or more required skills'
+                label='Skills'
+                description='Select one or more skills that could help to bring your idea to life'
```
LetsPoke, line 304: can we use Choose here also? 
so that we not using "Select" for skills and "Choose" for categories :)

LetsPoke (approved):

Issue #1472, opened 2023-06-06
BA: Hackathon descriptions are not pleasing


LetsPoke:
the hackaton descriptions on all pages are neithar eyecatching nor pleasing for the eyes 
we need to come up with a more pleasing visualization 

- look for examples in the internet 
- discuss with the others
- implement

PR #1513, opened 2023-06-14
#1472 ba: hackathon description visualisation changed

katty-bird:


hms-app/src/components/HackathonHeader.tsx
```diff
@@ -32,6 +32,7 @@ export default function HackathonHeader(props: IProps) {
               readOnly
               value={hackathonData.description || ''}
               id='hackathonDescriptionEditor'
+              style={{ color: 'gray', backgroundColor: 'transparent', border: 'none', textAlign: 'center' }}
```
LetsPoke, line 35: I think we should not align this to 'center' as we might have lists in the description and this could look weird

LetsPoke (approved):

PR #1514, opened 2023-06-14
fix: new version of button

reyhanrumengan:
Please also check the Dark mode version. 
Still not sure with it.

### DIFF
--- a/hms-app/src/components/buttons/ParticipateButton.tsx
+++ b/hms-app/src/components/buttons/ParticipateButton.tsx
@@ -1,4 +1,4 @@
-import { Group, Text, Switch } from '@mantine/core'
+import { Group, Text, Switch, useMantineTheme } from '@mantine/core'
 import {
   createIdeaParticipant,
   removeIdeaParticipant,
@@ -12,6 +12,8 @@ import { useMsal } from '@azure/msal-react'
 import { Check, X } from 'tabler-icons-react'
 import { Idea } from '../../common/types'
 import { styles } from '../../common/styles'
+import { IconCheck, IconX } from '@tabler/icons-react'
+import { JOIN_BUTTON_COLOR, LEAVE_BUTTON_COLOR } from '../../common/colors'
 
 type IProps = {
   idea: Idea
@@ -32,6 +34,7 @@ export default function ParticipateButton(props: IProps) {
     userId: '',
     participantId: '',
   })
+  const theme = useMantineTheme()
 
   const addHackathonParticipant = async (
     action = createIdeaParticipant,
@@ -184,9 +187,6 @@ export default function ParticipateButton(props: IProps) {
   return (
     <>
       <Group>
-        <Text className={classes.boldText}>
-          {participantCheck ? 'Leave' : 'Participate'}
-        </Text>
         <Switch
           disabled={buttonIsDisabled}
           checked={participantCheck}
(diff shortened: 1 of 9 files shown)


reyhanrumengan:
Still need to fix the design
Because when there's no "Join Idea" button, like in the `IdeaPool`, all of other buttons are on the most left side.

hms-app/src/components/buttons/ParticipateButton.tsx
```diff
@@ -181,18 +182,21 @@ export default function ParticipateButton(props: IProps) {
...
-      }}
-    >
-      {participantCheck ? 'Leave Idea' : 'Join Idea'}
-    </Button>
+    <>
+      <Group>
+        <Text className={classes.boldText}>
+          {participantCheck ? 'Leave' : 'Participate'}
+        </Text>
+        <Switch
```
LetsPoke, line 190: maybe we can use this kind of switch, so that we can always use the word "participate"

hms-app/src/components/buttons/ParticipateButton.tsx
```diff
@@ -181,18 +185,27 @@ export default function ParticipateButton(props: IProps) {
...
+          disabled={buttonIsDisabled}
+          checked={participantCheck}
+          onChange={
+            participantCheck
+              ? removeThisIdeaParticipant
+              : addThisIdeaParticipant
+          }
+          thumbIcon={
+            participantCheck ? (
+              <IconCheck size='0.8rem' color={JOIN_BUTTON_COLOR} stroke={3} />
```
LetsPoke, line 200: we could use the same "Check" and "X" compenent that is used in the notifications so we wouldnt need to load 2 libarys

hms-app/src/components/buttons/ParticipateButton.tsx
```diff
@@ -181,18 +185,27 @@ export default function ParticipateButton(props: IProps) {
...
-        backgroundColor: participantCheck
-          ? LEAVE_BUTTON_COLOR
-          : JOIN_BUTTON_COLOR,
-      }}
-    >
-      {participantCheck ? 'Leave Idea' : 'Join Idea'}
-    </Button>
+    <>
+      <Group>
+        <Switch
```
LetsPoke, line 190: could you add 
color={RELOAD_BUTTON_COLOR}
to the switch to make it green? ( I think green is more intuitiv then blue) 
and we should update the color file to 
export const RELOAD_BUTTON_COLOR = 'teal'

hms-app/src/components/buttons/ParticipateButton.tsx
```diff
@@ -1,4 +1,4 @@
-import { Button } from '@mantine/core'
+import { Group, Text, Switch, useMantineTheme } from '@mantine/core'
```
LetsPoke, line 1: Theme and text can be removed again I guess

hms-app/src/components/card-details/IdeaDetails.tsx
```diff
@@ -288,13 +288,31 @@ export default function IdeaDetails(props: IProps) {
...
     return (
-      <Text className={classes.smallText}>
-        Created: {new Date(idea.creationDate).toDateString()}
+      <Text mt={10} className={classes.smallText}>
+        Created: {formattedDate}
       </Text>
     )
   }
 
+  const buttons = () => {
```
LetsPoke, line 305: seems to be not used

LetsPoke:
Personal preference but what you think about it? 
Without mt=10 and with <Stack spacing='xs' style={{ paddingLeft: '5%' }}>
Bildschirm­foto 2023-06-16 um 13 50 33
with mt=10 and without spacing left 
Bildschirm­foto 2023-06-16 um 13 51 34

reyhanrumengan:
yeah, the first one looks better. I haven't fixed the spacing yet

LetsPoke (approved):

PR #1522, opened 2023-06-15
fix: add headline for hackathon selection

reyhanrumengan:


hms-app/src/pages/AllIdeas.tsx
```diff
@@ -111,7 +111,8 @@ function AllIdeas() {
         value={participantInfo.participantId}
       >
         <HackathonVotingContext.Provider value={hackathonData.votingOpened}>
-          <Group position={'apart'} my={20}>
+          <Text className={classes.title}>Hackathon Selection:</Text>
+          <Group position={'apart'} mb={20}>
             <HackathonSelectDropdown
```
LetsPoke, line 116: wouldnt it be better to add it to the HackathonSelectDropdown component? 
so that we dont need to add it to all pages extra

Issue #1433, opened 2023-05-30
feat: add search function to the ideation portal


LetsPoke:

LetsPoke:
and first fix it

PR #1533, opened 2023-06-20
#1433 feat add search function to the ideation portal

rz-p:


reyhanrumengan:
In the Archive, when I search for an idea and if the Idea that were shown only one, the search bar were moved to the center and the IdeaCard were moved to the right

Bildschirmfoto 2023-06-20 um 10 36 00

reyhanrumengan (approved):

LetsPoke:
Looks good 

is it possible to change the number of ideas to "searchIdea.length" when the searchTerm is not empty 
so that if 2 ideas are shown with seacht that the number of ideas is not saying 10

hms-app/src/pages/Archive.tsx
```diff
@@ -1,24 +1,26 @@
 import React, { useEffect, useState, useContext } from 'react'
 import HackathonSelectDropdown from '../components/HackathonSelectDropdown'
-import { Group, Text, Checkbox, Title, Stack } from '@mantine/core'
+import { Group, Text, Checkbox, Title, Stack, Input } from '@mantine/core'
```
LetsPoke, line 3: unused import

hms-app/src/pages/Archive.tsx
```diff
@@ -52,24 +54,28 @@ export default function Archive() {
     })
   }
 
-  const userIdea = relevantIdeaList.filter((item) => {
+  const handleChangeSearch = (event: React.ChangeEvent<HTMLInputElement>) => {
```
LetsPoke, line 57: unused const

LetsPoke (approved):

PR #1534, opened 2023-06-20
header update for idea pool and hackathons pages

katty-bird:
HackathonHeader.tsx
* removed badge with date
* applied heroHeaderStyles() to header on the hackathons page
* added formatHackathonDates() function to display hackathon dates in the format Jun 20, 2023 - Jun 20, 2023

AllIdeas.tsx
* HackathonHeader moved to the top of the page for uniform style of all pages
* increased distance between the group of buttons and the list of ideas

IdeaPool.tsx
* applied heroHeaderStyles() to header on the Idea pool page

### DIFF
--- a/hms-api/dev/init/hackathon-seed.json
+++ b/hms-api/dev/init/hackathon-seed.json
@@ -4,7 +4,7 @@
       "S": "e955fe4b-7ce7-4904-ae6f-22a8985f74a9"
     },
     "title": {
-      "S": "Ideation Portal"
+      "S": "Idea Pool"
     },
     "description": {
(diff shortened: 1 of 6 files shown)


LetsPoke:
few things: 
- when no hackathon is selected or loaded the header shows "invalid date - invalid date" (in that case the heade should not be visible)
- could you add the "new idea" and "join hackathon" button to the header? ( same way as the buttons are in the home screen )
- can we make the whole header to an accordion that is by default open but has a arrow in the middle of the bottom to "close" it

LetsPoke (approved):

